Issue #9, opened 2022-02-17
Reorganize the folders and the source files


fdiblen:
use correct file extensions: cpp, hpp
use include and src folder instead of cu
add example to `tests` folder

jspaaks:
Something like this:
```plain
.
├── cudawrappers
│   ├── CMakeLists.txt
│   ├── cudawrappers-config.cmake
│   ├── include
│   │   └── cudawrappers.hpp
│   ├── README.md
│   ├── src
│   │   └── cudawrappers.cpp
│   └── tests
│       ├── CMakeLists.txt
│       ├── convolution
│       │   ├── convolution.cpp
│       │   └── README.md
│       ├── README.md
│       └── vector_add
│           ├── README.md
│           ├── vector_add.cpp
│           └── vector_add.cu
└── cudawrappers-usage-examples
    ├── cmake-pull
    │   ├── CMakeLists.txt
    │   └── helloworld.cpp
    ├── git-submodule
    │   ├── cudawrappers@0.1.0
    │   └── helloworld.cpp
    ├── locally-installed
    │   ├── CMakeLists.txt
    │   └── helloworld.cpp
    ├── plain-copy
    │   ├── CMakeLists.txt
    │   ├── cudawrappers
    │   └── helloworld.cpp
    └── README.md
```

`cudawrappers-usage-examples` will be a separate repository

jspaaks:
Maybe use sagecal for inspiration on CMakeLists configuration

e.g. https://github.com/nlesc-dirac/sagecal/blob/master/CMakeLists.txt

fdiblen:
Suggestion from Abel: We should also add examples to the cudawrappers repo.

fdiblen:
@benvanwerkhoven @isazi @bouweandela What do you think about the proposed folder structure?

bouweandela:
@fdiblen Looks good to me, except that I'm not super happy with the README.md files buried in the directory tree: this makes information hard to find. Why not have all info in the README.md at the top level or on a documentation webpage?

It would also be good to hear the opinion of our project partners @john-romein @csbnw and @svlugt on this. Does this work for you?

jspaaks:
PR #43 was merged, I think we can close this issue. If we feel Bouwe's comment about READMEs above should be addressed, we should make a separate issue for it.

ping @fdiblen

PR #43, opened 2022-02-21
restructure file tree

jspaaks:
In this PR:

- restructure the file tree according to the layout in #9, it leaves the file content of all files unchanged (we'll work on that later)
- rename `*.cc` files to `*.cpp`
- rename `*.h` files to `*.hpp`
- added 4 usage directories. These are examples of how users may include the library:
    1. by including the library as a plain copy (`plain-copy/`)
    2. by includign the library as a copy made with git submodules (`git-submodule/`)
    3. by locally installing the library to their system (`locally-installed/`)
    4. by using cmake to collect a copy of the dependency from the internets (`cmake-pull/`)
- added cudawrappers config cmake file, to help users use this library in their cmake files without hassle

PS we can still split off the usage examples subtree later if we feel that's more clear for consumers.

Refs: 

- #9

How to check:

- look at directory structure online or in locally checked out copy of the repo


fdiblen (approved):
Thanks @jspaaks 
The current changes breaks the build. To move forward I will create a separate issue to fix the build system.

jspaaks:
Thanks for reviewing!

Issue #23, opened 2022-02-17
Use a static code analysis tool


fdiblen:
We need a tool that can analyse the code and generate a report.
https://sonarcloud.io/ can be one of the options.

Ask @benvanwerkhoven and @isazi for suggestions.

isazi:
Sonarcloud supports C/C++ applications, not sure about CUDA.

benvanwerkhoven:
Yeah I have reasonably good experience using sonarcloud on the Kernel Tuner repository. The only issue I have is that it complains about using pseudo-random numbers being insecure, while I use those for stochastic optimization in a context unrelated to security. But after many attempts I seem to have been able to disable most of those checks now, I think.

isazi:
And I believe it is possible to disable checks in the libraries.

jspaaks:
It seems everybody is on board with SonarCloud. @fdiblen I suggest updating the issue title to _Set up sonarcloud_ so this issue becomes more actionable.

bouweandela:
Is it possible to run the Sonarcloud linter(s) locally, e.g. from a pre-commit hook? Or is it a proprietary linter?

bouweandela:
Maybe this list could also be used for inspiration? https://github.com/caramelomartins/awesome-linters#cc

bouweandela:
Related discussion in #26

abelsiqueira:
I had issues installing SonarCloud, but it looks like it is working now.

Looks like @loostrum made a request to enable it for the organization in the past, and it conflicted with my request. Since there were no projects already using it for this organization, I uninstalled it and started from scratch. @loostrum, I have not enabled it for your repo, but it should be possible to do it now.

bouweandela:
@csbnw mentioned that at Astron they make use of:
- C++ code: `clang-format` and `clang-tidy` with [Google code style](https://google.github.io/styleguide/cppguide.html) 
- Cmake files: [`cmake-format` and `cmake-lint`](https://cmake-format.readthedocs.io/en/latest/)

fdiblen:
We will look into codacy as it seems to be supporting the tools we use locally. See https://docs.codacy.com/getting-started/supported-languages-and-tools/

abelsiqueira:
PR merged.

Issue #32, opened 2022-02-17
Write developer documentation on running linters locally


jspaaks:

fdiblen:
one option is https://github.com/danmar/cppcheck

abelsiqueira:
- [ClangFormat](https://clang.llvm.org/docs/ClangFormat.html)
- Maybe more from https://github.com/fffaraz/awesome-cpp#static-code-analysis

abelsiqueira:
Following #23, here's a short summary so I don't work on this issue at the same time:

Install
- `cmake-format` (https://cmake-format.readthedocs.io/en/latest/installation.html)
- VSCode extension for `cmake-format` (https://marketplace.visualstudio.com/items?itemName=cheshirekow.cmake-format)
- `clang-tidy` (https://clang.llvm.org/extra/clang-tidy/)
- VSCode extension `clang-tidy` (https://marketplace.visualstudio.com/items?itemName=notskm.clang-tidy)
- `clang-format` (https://clang.llvm.org/docs/ClangFormat.html)
- VSCode extension Clang-Format (https://marketplace.visualstudio.com/items?itemName=xaver.clang-format)

abelsiqueira:
#59 has the commands.

abelsiqueira:
PR was merged.

PR #59, opened 2022-02-24
Add linters

abelsiqueira:
**Description**

Add linters. Originally I was planning to use [Schaapcommon](https://git.astron.nl/RD/schaapcommon), but they have a GPL-3.0 license, and we have Apache 2.0. Instead, I will use basic or default files.

**Related issues**:

- #23 

Clone and verify that the linters run. There should be errors and modifications because I don't run or format anything on this PR.

Install [clang-format](https://clang.llvm.org/docs/ClangFormat.html), [clang-tidy](https://clang.llvm.org/extra/clang-tidy/), Python package [cmakelang](https://cmake-format.readthedocs.io/en/latest/installation.html), [cppcheck](https://cppcheck.sourceforge.io) and [flawfinder](https://dwheeler.com/flawfinder/)
```
git clone https://github.com/nlesc-recruit/CUDA-wrappers .
git checkout <this-branch>
cmake -S . -B build
make -C build format
# compare to running individually
make -C build clang-format
make -C build cmake-format

make -C build lint
# compare to running individually
make -C build clang-tidy
make -C build cmake-lint
make -C build cppcheck
make -C build flawfinder
```

### DIFF
--- a/.github/workflows/codacy-analysis.yml
+++ b/.github/workflows/codacy-analysis.yml
@@ -1,0 +1,28 @@
+name: Codacy Analysis CLI
+
+on: ["push"]
+
+jobs:
+  codacy-analysis-cli:
+    name: Codacy Analysis CLI
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@main
+
+      - name: Install requirements
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y clang-tidy nvidia-cuda-toolkit
+
+      - name: Run clang-tidy and codacy-clang-tidy
+        env:
+          CODACY_CLANG_TIDY_VERSION: 1.3.2
+          PROJECT_TOKEN: ${{ secrets.CODACY_PROJECT_TOKEN }}
+        run: |
+          export COMMIT=$(if [ ${{ github.event_name }} == "pull_request" ]; then echo "${{ github.event.pull_request.head.sha }}"; else echo "${{ github.sha }}"; fi)
+          wget https://github.com/codacy/codacy-clang-tidy/releases/download/${CODACY_CLANG_TIDY_VERSION}/codacy-clang-tidy-linux-${CODACY_CLANG_TIDY_VERSION} -O codacy-clang-tidy-latest
+          git clone https://github.com/codacy/codacy-clang-tidy
+          chmod a+x codacy-clang-tidy-latest codacy-clang-tidy/scripts/send-results.sh
+          cmake -S . -B build
+          make -C build clang-tidy | codacy-clang-tidy/scripts/send-results.sh
--- a/.github/workflows/codacy.yml
+++ b/.github/workflows/codacy.yml
@@ -1,20 +1,22 @@
-name: Codacy Analysis CLI
+name: Codacy
 
 on: ["push"]
 
 jobs:
-  codacy-analysis-cli:
-    name: Codacy Analysis CLI
+  codacy:
+    name: Codacy
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
-        uses: actions/checkout@main
+        uses: actions/checkout@v2
 
       - name: Install requirements
         run: |
           sudo apt-get update
           sudo apt-get install -y clang-tidy nvidia-cuda-toolkit
 
+      # COMMIT variable defined below was taken from codacy-analysis-cli-action
+      #   https://github.com/codacy/codacy-analysis-cli-action/blob/9f09141804faa25deeb6d09264f7e894f6c50ef9/action.yml#L92
       - name: Run clang-tidy and codacy-clang-tidy
         env:
           CODACY_CLANG_TIDY_VERSION: 1.3.2
@@ -25,4 +27,4 @@ jobs:
           git clone https://github.com/codacy/codacy-clang-tidy
           chmod a+x codacy-clang-tidy-latest codacy-clang-tidy/scripts/send-results.sh
           cmake -S . -B build
-          make -C build clang-tidy | codacy-clang-tidy/scripts/send-results.sh
+          make --directory=build clang-tidy | codacy-clang-tidy/scripts/send-results.sh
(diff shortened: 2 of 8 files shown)


jspaaks:
Hi Abel, nice work on this. What do you think about adding the install instructions for cmakelang, clang-tidy, clan-format to the developer documentation `README.dev.md` following Bouwe's suggestion?

abelsiqueira:
> Hi Abel, nice work on this. What do you think about adding the install instructions for cmakelang, clang-tidy, clan-format to the developer documentation `README.dev.md` following Bouwe's suggestion?

Issue #32 touches that already.

bouweandela:
Why do you need so many configuration files if you're using the defaults?

abelsiqueira:
> Why do you need so many configuration files if you're using the defaults?

So the choices are explicit. Otherwise, the software will look for configuration files in whatever other places and that can lead to different results. The commands to generate these are in the files.

fdiblen (changes requested):
thank you @abelsiqueira 

When I run `make -C build clang-lint` I get the error below.

```
make: Entering directory '/tmp/tmp.K30e8g7bzm/build'
make: *** No rule to make target 'clang-lint'.  Stop.
make: Leaving directory '/tmp/tmp.K30e8g7bzm/build'
```

Maybe we are missing a configuration for the linter?

abelsiqueira:
Nice catch. Apparently, this is a bug with empty CMakeLists.txt in `usage-examples`. This can be verified with
```
cmake-lint usage-examples/plain-copy/CMakeLists.txt
```
While for the root CMakeLists.txt it will work normally. Notice that `cmake-format` fixes the empty CMakeLists.txt to have a single empty line, so running these in order, like I did, does not reveal this error.

abelsiqueira:
> Including the generated files has the downside that changes to the default configuration will not be picked up as the linter/formatter is updated. Is that intentional? Or would it be better to e.g. just tell clang-format to use Google style instead of the Google style dumped to a configuration file?

Intentional. If I have a different version than you and we rely on the versions' default then we have different results. Also, I expected that ASTRON has some specific style, so this can also be seen as a placeholder for their choices.

.github/workflows/codacy-analysis.yml
```diff
@@ -0,0 +1,28 @@
+name: Codacy Analysis CLI
+
+on: ["push"]
+
+jobs:
+  codacy-analysis-cli:
```
fdiblen, line 6:
```suggestion
  codacy:
```

.github/workflows/codacy-analysis.yml
```diff
@@ -0,0 +1,28 @@
...
+    name: Codacy Analysis CLI
```
fdiblen, line 7:
```suggestion
    name: Codacy
```

.github/workflows/codacy.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Run clang-tidy and codacy-clang-tidy
+        env:
+          CODACY_CLANG_TIDY_VERSION: 1.3.2
+          PROJECT_TOKEN: ${{ secrets.CODACY_PROJECT_TOKEN }}
+        run: |
+          export COMMIT=$(if [ ${{ github.event_name }} == "pull_request" ]; then echo "${{ github.event.pull_request.head.sha }}"; else echo "${{ github.sha }}"; fi)
+          wget https://github.com/codacy/codacy-clang-tidy/releases/download/${CODACY_CLANG_TIDY_VERSION}/codacy-clang-tidy-linux-${CODACY_CLANG_TIDY_VERSION} -O codacy-clang-tidy-latest
+          git clone https://github.com/codacy/codacy-clang-tidy
+          chmod a+x codacy-clang-tidy-latest codacy-clang-tidy/scripts/send-results.sh
+          cmake -S . -B build
```
abelsiqueira, line 29: These options do not have long versions

.github/workflows/codacy.yml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Install requirements
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y clang-tidy nvidia-cuda-toolkit
+
+      # COMMIT variable defined below was taken from codacy-analysis-cli-action
```
abelsiqueira, line 18: Explaining where did `COMMIT` came from.

cmake/linter-tools.cmake
```diff
@@ -0,0 +1,47 @@
...
+
+# Ref: https://stackoverflow.com/questions/32280717/cmake-clang-tidy-or-other-script-as-custom-target
+
+file(GLOB_RECURSE ALL_SOURCE_FILES *.cpp *.hpp)
+file(GLOB_RECURSE ALL_CMAKE_LISTS CMakeLists.txt)
+
+add_custom_target(
+    clang-format
+    COMMAND clang-format
+    -i
```
abelsiqueira, line 12: This does not have a long form.

cmake/linter-tools.cmake
```diff
@@ -0,0 +1,47 @@
...
+)
+
+set(INCLUDE_DIRECTORIES -I../include/ -I${CUDAToolkit_INCLUDE_DIRS})
+add_custom_target(
+    clang-tidy
+    COMMAND clang-tidy
+    ${ALL_SOURCE_FILES}
+    --
+    -std=c++11
+    ${INCLUDE_DIRECTORIES}
```
abelsiqueira, line 23: `--` separates options for `clang-tidy` and options for the compiler.
Maybe `INCLUDE_DIRECTORIES` can be reused in the compilation of the rest of the library, but it is probably best discussed in a different issue.

abelsiqueira:
@fdiblen, I looked into the `clang-tidy` variables, and it needs to build the library to work properly. That is why it needs `INCLUDE_DIRECTORIES`. However, I did not set INCLUDE_DIRECTORIES correctly.
At the moment, I am still not sure that Codacy + Clang-tidy is working properly, but I hope the other issues were addressed. Please let me know.

fdiblen (approved):
Looks good. Thanks @abelsiqueira :stew:

abelsiqueira:
Thanks for the review :rice:

Issue #27, opened 2022-02-17
Add cffconvert-github-action


jspaaks:

jspaaks:
PR was merged, closing this issue

PR #60, opened 2022-02-28
Added cffconvert workflow to validate CITATION.cff

jspaaks:
**Description**

Add the cffconvert github action to validate CITATION.cff in CI

See https://github.com/marketplace/actions/cffconvert

**Related issues**:

- #27

Review online or using the manual dispatch in Github Actions after merge.

### DIFF
--- a/.github/workflows/cffconvert.yml
+++ b/.github/workflows/cffconvert.yml
@@ -1,0 +1,19 @@
+name: cffconvert
+
+on:
+  push:
+    paths:
+      - CITATION.cff
+
+jobs:
+  validate:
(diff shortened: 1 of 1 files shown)


abelsiqueira (approved):
I was thinking of something like "Add CITATION.cff validator", but maybe you're right that it doesn't need to go in the CHANGELOG. I am approving so you can decide either way.

jspaaks:
OK, I created #67 for the time being so we can keep moving forward

jspaaks:
Thanks for reviewing and the discussion!

Issue #45, opened 2022-02-21
Update .gitignore


abelsiqueira:
Namely:
- `build`
- `CMakeFiles`

Possibly other things.

jspaaks:
Found this, may be helpful

- gist https://github.com/kigster/cmake-project-template/blob/8c583cafb7d5101098746bcef9a5bcfd3c5cd565/.gitignore
- Kitware's own gitignore https://gitlab.kitware.com/cmake/cmake/-/blob/master/.gitignore not sure why they have pyc in there

Also, CMake tip about arbitrarily named build dirs might be useful

- https://www.scivision.dev/cmake-auto-gitignore-build-dir/

fdiblen:
Fixed by #80

PR #80, opened 2022-03-01
Update gitignore

fdiblen:
**Description**

- Updates .gitignore to ignore generated CMake artifacts, binaries and editor configurations.
- Updates CMakeLists.txt to copy .ignore to build directory (thanks for the tip @jspaaks )

**Related issues**:

- #45 

- Review the list in .gitignore file ;)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,5 @@
-# Prerequisites
-*.d
-
 # Compiled Object files
+**/.DS_Store
 *.slo
 *.lo
 *.o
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,7 +1,5 @@
-# Prerequisites
-*.d
-
 # Compiled Object files
+**/.DS_Store
```
jspaaks, line 2: Is this a "Compiled Object file"? I  thought it was just a cache file that's used on Macs

.gitignore
```diff
@@ -30,3 +28,39 @@
...
+**/*.cbp
+**/CMakeScripts
+**/compile_commands.json
+**/CMakeLists.txt.user
+**/_deps
+**/Testing
+**/CMakeUserPresets.json
+
+# Local 
+build/**/*
```
jspaaks, line 49: The `build/` thing here is a bit of a duplication of the CMake trick right? But might still be useful for people who want to not use CMake for some reason

jspaaks (approved):
Good to go, but I left 2 comments you may find interesting.

.gitignore
fdiblen: Now called generated files

fdiblen:
> Good to go, but I left 2 comments you may find interesting.

Thanks for reviewing @jspaaks :shallow_pan_of_food:

Issue #116, opened 2022-03-03
Add build script for vector_add


abelsiqueira:

abelsiqueira:
I have reached a point of being able to compile the example (#123), but when I try to run it on DAS-5, I get

```
cu::Error: named symbol not found
```

So I went back to version 0.1.0 and tried to compile the code before any of the major changes, but I still get the same error. Maybe someone with more C++ experience can help out? cc. @benvanwerkhoven @isazi @csbnw @john-romein @svlugt 

Steps to reproduce on DAS-5, no CMake, no new changes:

```
module load cuda10.1
git clone https://github.com/nlesc-recruit/cudawrappers
cd cudawrappers
git checkout 0.1.0
nvcc cu/*.cc examples/vector_add/vector_add.cc examples/vector_add/vector_add_kernel.cu -o vector_add -I. -lcuda -lnvrtc
alias gpurun='srun -N 1 -C TitanX --gres=gpu:1'
gpurun vector_add
```

benvanwerkhoven:
Looks to me like you are not linking the cudawrappers library. 

The Makefile that I used to compile the vector_add example is:

```
LIBS = -L../../build/ -lcuda -lnvrtc


CUDA_WRAPPER_INCLUDE = -I../../
CUDA_WRAPPER_LIB = -Wl,-rpath=../../build/ -lcu



all: vector_add


vector_add: vector_add.cc vector_add_kernel.cu
	g++ $(LIBS) $(CUDA_WRAPPER_LIB) vector_add.cc $(CUDA_WRAPPER_INCLUDE) -o vector_add


clean:
	rm vector_add
```

abelsiqueira:
> Looks to me like you are not linking the cudawrappers library.

I added the source files explicitly (See my command `nvcc cu/*.cc ...`) to remove CMake from the issue, so that was not it.

But your answer showed me the solution. Apparently, the `.cu` is not necessary **but** it needs to be in the same folder.
This works now:

```shell
nvcc ../../cu/*.cc vector_add.cc -o vector_add -I../../ -lcuda -lnvrtc
gpurun vector_add
```

abelsiqueira:
@benvanwerkhoven, do you know the proper way of handling the issue of having `.cu` in the same folder?

benvanwerkhoven:
[...]

Issue #89, opened 2022-03-01
check if clang-tidy works


fdiblen:
clang-tidy was introduced in #59 but we are not sure if it actually does something.

abelsiqueira:
Does not work: https://github.com/nlesc-recruit/cudawrappers/runs/5392442336?check_suite_focus=true#step:4:336

abelsiqueira:
It is working now.

PR #117, opened 2022-03-03
Update clang-tidy configuration and lint only src/ and include/ files

abelsiqueira:
**Description**

First, I updated clang-tidy's configuration.
The old configuration essentially ran no (significant) checks. The new one was taken from https://github.com/fbaeuerlein/cpp-vscode-guide (CC0 License).

This may lead to differences between clang-format and clang-lint, but this is for another issue.

One alternative to using the linked repo is to use `clang-tidy -checks '*' -dump-config > .clang-tidy` and use **all** checks. I chose not to do it because the configuration will have 650 lines and tens of thousands of errors are returned by clang-tidy.

Second, `clang-tidy` needs to be able to compile the code, so I am linting only `src` and `include` files because the files in `tests` fail to build. #116 needs to be done beforing linting `tests` with `clang-tidy`.

**Related issues**:

- #89

Clone and verify
```
git clone https://github.com/nlesc-recruit/cudawrappers .
git checkout <this-branch>
cmake -S . -B build
make -C build format # Nothing should change
make -C build lint # linting is broken until we fix the issues (see #92)
```

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,47 +1,38 @@
 ---
-# Generated with clang-tidy src/* -dump-config > .clang-tidy
-Checks:          'clang-diagnostic-*,clang-analyzer-*'
-WarningsAsErrors: ''
+# Copied from https://github.com/fbaeuerlein/cpp-vscode-guide/blob/02617fe158f456d254fbebffcf6a7957198cbc5f/.clang-tidy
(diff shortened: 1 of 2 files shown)


fdiblen (approved):
LGTM. :meat_on_bone:

abelsiqueira:
Thanks for the review :fish:

PR #144, opened 2022-04-06
Fix include in CMakeLists.txt

### LINKED ISSUE #134: CMakeLists uses legacy commands that make composition by consumers more difficult
Relevant code https://github.com/nlesc-recruit/cudawrappers/blob/8145a61d2aa2adc087ee1a74a57028d45a72fd16/CMakeLists.txt#L21-L22

CMake docs says to not use `include_directories`: https://cmake.org/cmake/help/v3.17/command/include_directories.html

Starting approximately in CMake 3.0, the building paradigm has been moved to describe build "targets", to which you can attach information on how their information should propagate to other elements. This makes it easier to compose larger projects with multiple dependencies, without having them infere with one another.

bouweandela:
**Description**

Update the include statement in CMakeLists.txt so this can be used in the git submodule example.

**Related issues**:

Closes #134

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,8 @@ project(
   DESCRIPTION "C++ Wrappers for the CUDA Driver API"
   VERSION 0.3.0
   HOMEPAGE_URL https://github.com/nlesc-recruit/cudawrappers
-  LANGUAGES CUDA CXX)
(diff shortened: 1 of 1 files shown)


bouweandela:
Closed in favour of #150.

PR #151, opened 2022-04-14
Add cufft library

### LINKED ISSUE #146: Add cufft
Code lives here: https://github.com/nlesc-recruit/cudawrappers/blob/0.1.0/cu/cufft.cc

bouweandela:
**Description**

Add cufft library

**Related issues**:

Closes #146

- Check that the new functionality can be used


csbnw:
There are many Codacy suggestions for the code (the `cufft.*` files), these should either be addressed in this PR, or in a new issue.

include/cufft.hpp
```diff
@@ -0,0 +1,103 @@
...
+template <>
+inline FFT<std::complex<half>, std::complex<half>, 1>::FFT(unsigned n,
+                                                           unsigned count) {
+  checkCuFFTcall(cufftCreate(&plan));
+
+  long long size = n;
+  size_t neededWorkSize{};
+  checkCuFFTcall(cufftXtMakePlanMany(plan, 1, &size, nullptr, 1, 1, CUDA_C_16F,
+                                     nullptr, 1, 1, CUDA_C_16F, count,
+                                     &neededWorkSize, CUDA_C_16F));
```
csbnw, line 99: I don't think it's correct to use hard-coded `CUDA_C_16F` here.

include/cufft.hpp
```diff
@@ -0,0 +1,103 @@
...
+  const char *what() const noexcept override;
+
+  operator cufftResult() const { return _result; }
+
+ private:
+  cufftResult _result;
+};
+
+template <typename Tin = cufftComplex, typename Tout = cufftComplex,
+          unsigned DIM = 1>
```
csbnw, line 27: In it's current state, it doesn't seem like 2D FFTs are supported.

src/cufft.cpp
```diff
@@ -0,0 +1,30 @@
...
+    "invalid device",
+    "parse error",
+    "no workspace",
+    "not implemented",
+    "license error",
+    "not supported"};
+
+const char *Error::what() const noexcept {
+  return CUFFT_STATUS_MSG.at(_result);
+}
```
csbnw, line 28: Isn't it safer to use an enum here, like [so](https://github.com/NVIDIA/cuda-samples/blob/b312abaa07ffdc1ba6e3d44a9bc1a8e89149c20b/Common/helper_cuda.h#L108)?

john-romein: It used to be implemented as a large switch statement, but this is not strictly necessary as NVIDIA documents the error numbers (not just the names) in the manual.

bouweandela: One of the linters complained about that, that's why I changed it. I can't quite remember what the complaint was though, since it is some time ago.

tests/convolution/convolution.cpp
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  return ArrayComparisonStats::computeStats(totalDiff, exceedingCount,
+                                            totalElements);
+}
+
+int main(int argc, char *argv[]) {
+  const unsigned fftSize = 1024u;
+  const size_t arraySize = sizeof(cufftComplex) * fftSize * fftSize;
+  std::cout << "HERE WE ARE";
```
csbnw, line 74:
```suggestion
```

tests/convolution/convolution.cpp
```diff
@@ -0,0 +1,104 @@
...
+      ofile.write((char *)&array[k].y, sizeof(float));
+    }
+    ofile.close();
+    std::cout << "Completed" << std::endl;
+  } else {
+    std::cerr << "Cannot open file " << fname << std::endl;
+    exit(1);
+  }
+}
+
```
csbnw, line 23: This function isn't used (anymore), please consider removing it:
```suggestion
```

tests/convolution/convolution.cpp
```diff
@@ -0,0 +1,104 @@
...
+  const size_t arraySize = sizeof(cufftComplex) * fftSize * fftSize;
+  std::cout << "HERE WE ARE";
+  cu::init();
+  cu::Device device(0);
+  cu::Context context(CU_CTX_SCHED_BLOCKING_SYNC, device);
+  cu::DeviceMemory in_dev(arraySize), out_dev(arraySize);
+  cu::HostMemory in_host(arraySize), out_host(arraySize),
+      out_testing(arraySize);
+
+  cu::Stream my_stream;
```
csbnw, line 82: Rename to `stream`, like in the other examples/tests?
```suggestion
  cu::Stream stream;
```

include/cufft.hpp
csbnw: This is fixed in the latest version.

tests/CMakeLists.txt
```diff
@@ -16,15 +16,16 @@ include(Catch)
...
-set(COMPONENTS cu nvrtc)
+set(COMPONENTS cu nvrtc cufft)
 foreach(component ${COMPONENTS})
-  add_executable(test_${component})
+  add_executable(test_${component} test_${component}.cpp)
   target_sources(test_${component} PRIVATE test_${component}.cpp)
-  target_link_libraries(
-    test_${component} PUBLIC cudawrappers::${component} tests
-  )
+  target_link_libraries(test_${component} PUBLIC ${LINK_${component}} tests)
```
csbnw, line 28: @bouweandela , you changed these lines, but I am not sure why this was needed specifically for the new `test_cufft`.

matmanc: I think because more then one library is used by cufft in particular it needs both cu an cufft

include/cudawrappers/cufft.hpp
```diff
@@ -0,0 +1,223 @@
...
+ */
+class FFT {
+ public:
+  FFT() = default;
+  FFT &operator=(FFT &) = delete;
+  FFT(FFT &) = delete;
+  FFT &operator=(FFT &&other) noexcept {
+    if (&other != this) {
+      plan_ = other.plan_;
+      other.plan_ = 0;
```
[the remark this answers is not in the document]
csbnw: According to `cufft.h`:
```
// cufftHandle is a handle type used to store and access CUFFT plans.
typedef int cufftHandle;
```
So `0` as it currently is seems appropriate.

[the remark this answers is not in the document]
matmanc: In the docs it says https://docs.nvidia.com/cuda/cufft/index.html?highlight=cufftHandle#cuffthandle
typedef unsigned int cufftHandle;
weird

examples/vector_add/CMakeLists.txt
```diff
@@ -14,7 +14,9 @@ function(include_cuda_code target input_file)
 endfunction(include_cuda_code)
 
 add_executable(vector_add vector_add.cpp)
-target_link_libraries(vector_add PRIVATE cudawrappers::cu cudawrappers::nvrtc)
+target_link_libraries(
+  vector_add PRIVATE cudawrappers::cu CUDA::cudart cudawrappers::nvrtc
+)
```
loostrum, line 19: Is there a reason cudart is now linked here? The vector_add example runs fine on the main branch as well

include/cudawrappers/cufft.hpp
```diff
@@ -0,0 +1,223 @@
...
+
+#include <exception>
+
+#include "cudawrappers/cu.hpp"
+
+namespace {
+/*
+ * Error handling helper function, copied from cuda-samples Common/helper_cuda.h
+ */
+static const char *_cudaGetErrorEnum(cufftResult error) {
```
loostrum, line 16: Is it necessary to have both the anonymous namespace and `static` declaration? I'm not too familiar with using them like this, but as far as I know they have the same purpose.

tests/CMakeLists.txt
```diff
@@ -16,15 +16,16 @@ include(Catch)
...
+set(LINK_cu cudawrappers::cu)
+set(LINK_cufft cudawrappers::cu cudawrappers::cufft)
+set(LINK_nvrtc cudawrappers::nvrtc)
 
 # Add library tests
-set(COMPONENTS cu nvrtc)
+set(COMPONENTS cu nvrtc cufft)
 foreach(component ${COMPONENTS})
-  add_executable(test_${component})
+  add_executable(test_${component} test_${component}.cpp)
```
loostrum, line 26: Isn't the addition of the cpp file here duplicate with the next line?

tests/test_cufft.cpp
```diff
@@ -0,0 +1,215 @@
+
```
loostrum, line 1: This first empty line should be removed.

loostrum (changes requested):
I've left some minor comments, but overall it looks good to me! I'm not very familiar with cufft itself, so I'm assuming that the actual functionality is correct. Glad to see the tests passing at least. In addition to the few requested changes, could you also update the changelog file?

examples/vector_add/CMakeLists.txt
csbnw: I am not sure why/when this was introduced, but I will revert it.

include/cudawrappers/cufft.hpp
csbnw: It indeed seems like we don't need the anonymous namespace here.

tests/test_cufft.cpp
csbnw: :heavy_check_mark:

loostrum (approved):

PR #159, opened 2022-05-05
Improve error message if file with cuda code does not exist

bouweandela:
This improves the error message if the file with cuda code cannot be opened when constructing an `nvrtc::Program`. Currently it is
```
cu::Error: named symbol not found
```
This pull request changes it to
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Error opening file: 'vector_add_kernel.cu'
```

### DIFF
--- a/include/nvrtc.hpp
+++ b/include/nvrtc.hpp
@@ -42,7 +42,8 @@ class Program {
   Program(const std::string &filename) {
     std::ifstream ifs(filename);
     if (!ifs.is_open()) {
-      throw std::runtime_error("Error opening file: '" + filename + "'");
+      throw std::runtime_error("Error opening file '" + filename +
+                               "' in cudawrappers::nvrtc");
     }
     std::string source(std::istreambuf_iterator<char>{ifs}, {});
     checkNvrtcCall(nvrtcCreateProgram(&program, source.c_str(),
--- a/include/nvrtc.hpp
+++ b/include/nvrtc.hpp
@@ -42,8 +42,7 @@ class Program {
   Program(const std::string &filename) {
     std::ifstream ifs(filename);
     if (!ifs.is_open()) {
-      throw std::runtime_error(
-          std::string("Error opening file: '" + filename + "'"));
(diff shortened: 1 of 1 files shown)


bouweandela:
Or would it be better to throw an `nvrtc::Error`? In that case, maybe derive `nvtrtc::Error` from `std::runtime_error` so it can be constructed from a `std::string`?

csbnw:
> Or would it be better to throw an `nvrtc::Error`?
I like `nvrtc:Error`, it immediately clarifies that something went wrong during runtime-compilation (with NVRTC).

john-romein:
I independently ran into the same problem.  I solved it by setting ifs.exceptions(std::ifstream::failbit | std::ifstream::badbit); but I think either solution is fine.  Throwing some nvrtc::Error would also be an option, but there is no nvrtcResult error code for a file that cannot be opened.

bouweandela:
>  I solved it by setting ifs.exceptions(std::ifstream::failbit | std::ifstream::badbit);

That's what I tried to do first too, but then the error message becomes
```
terminate called after throwing an instance of 'std::__ios_failure'
  what():  basic_ios::clear: iostream error
srun: error: node502: task 0: Aborted (core dumped)
```
which I didn't find very clear either, even if it's an improvement over the previous message.

>  Throwing some nvrtc::Error would also be an option, but there is no nvrtcResult error code for a file that cannot be opened.

Good point. If a one-to-one relation between `nvrtc::Error` and `nvrtcResult` is desired, that is not feasible. Let's keep the code as it is then.

svlugt (approved):
Tested:

When the file is in the path:
```
~/usage-example-locally-installed $ vector_add_example
hurray! 2
```

When the file is not in the path:
~/usage-example-locally-installed $ cd ..
```
~ $ vector_add_example
terminate called after throwing an instance of 'std::runtime_error'
  what():  Error opening file 'vector_add_kernel.cu' in cudawrappers::nvrtc
Aborted (core dumped)
```

PR #158, opened 2022-04-26
Address linter issues in cu and nvrtc libraries

bouweandela:
**Description**

Address linter issues in cu and nvrtc libraries

**Related issues**:

Fixes part of #92

**Instructions to review the pull request**

- Check that things still work as expected
- Check that no linter issues are reported anymore

PR #168, opened 2022-07-06
Minor changes needed for integration into DP3

csbnw:
**Description**
To integrate cudawrappers into [DP3](https://git.astron.nl/RD/DP3), a few missing methods had to be added.

### DIFF
--- a/include/cu.hpp
+++ b/include/cu.hpp
@@ -415,6 +415,11 @@ class Stream : public Wrapper<CUstream> {
 
   Stream(CUstream stream) : Wrapper<CUstream>(stream) {}
 
+  void memcpyHtoHAsync(void *dstPtr, const void *srcPtr, size_t size) {
+    checkCudaCall(cuMemcpyAsync(reinterpret_cast<CUdeviceptr>(dstPtr),
+                                reinterpret_cast<CUdeviceptr>(srcPtr), size, _obj));
+  }
+
   void memcpyHtoDAsync(CUdeviceptr devPtr, const void *hostPtr, size_t size) {
     checkCudaCall(cuMemcpyHtoDAsync(devPtr, hostPtr, size, _obj));
   }
--- a/include/cu.hpp
+++ b/include/cu.hpp
@@ -252,7 +252,7 @@ class DeviceMemory : public Wrapper<CUdeviceptr> {
 
   void zero(size_t size);
 
-  void zero(size_t size, Stream& stream);
+  void zero(size_t size, Stream &stream);
 
   const void *parameter()
       const  // used to construct parameter list for launchKernel();
@@ -421,7 +421,8 @@ class Stream : public Wrapper<CUstream> {
 
   void memcpyHtoHAsync(void *dstPtr, const void *srcPtr, size_t size) {
     checkCudaCall(cuMemcpyAsync(reinterpret_cast<CUdeviceptr>(dstPtr),
-                                reinterpret_cast<CUdeviceptr>(srcPtr), size, _obj));
+                                reinterpret_cast<CUdeviceptr>(srcPtr), size,
+                                _obj));
   }
 
   void memcpyHtoDAsync(CUdeviceptr devPtr, const void *hostPtr, size_t size) {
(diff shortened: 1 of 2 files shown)


csbnw:
@bouweandela , The pre-commit check is failing due to the code that was not touched in this PR. The static analysis doesn't like `reinterpret_cast`. There are C++20 [alternatives](https://rules.sonarsource.com/cpp/RSPEC-3630), but I am not sure what standard cudawrappers actually uses. How to handle this?

include/cu.hpp
```diff
@@ -423,6 +433,10 @@ class Stream : public Wrapper<CUstream> {
     checkCudaCall(cuMemcpyDtoHAsync(hostPtr, devPtr, size, _obj));
   }
 
+  void memcpyDtoDAsync(CUdeviceptr dstPtr, CUdeviceptr srcPtr, size_t size) {
```
bouweandela, line 436: Why are `dstPtr` and `srcPtr` of type `CUdeviceptr` instead of `DeviceMemory`, our wrapped version of `CUdeviceptr`?

bouweandela:
> I am not sure what standard cudawrappers actually uses

Currently C++11:
https://github.com/nlesc-recruit/cudawrappers/blob/dee4224bf7423406d98a035310c3e7d1eb7c3db3/CMakeLists.txt#L11
but if you and @john-romein and @svlugt think we could move to a newer standard we can update it?

bouweandela:
>  The pre-commit check is failing due to the code that was not touched in this PR. 

I know, I'm working on it in #158

include/cu.hpp
```diff
@@ -250,6 +250,10 @@ class DeviceMemory : public Wrapper<CUdeviceptr> {
     checkCudaCall(cuMemHostGetDevicePointer(&_obj, hostMemory, 0));
   }
 
+  void zero(size_t size);
```
bouweandela, line 253:
```suggestion
  void zero(const size_t size);
```
Would it make sense to do this? Or does that cause trouble?

john-romein: A const before a parameter of some basic type does not do anything useful, does it?  The parameter is passed by value, so from the function user's perspective, the value will not change anyway, with or without a const.

bouweandela: Good point, but I think it would protect you from you accidentally changing the parameter in the function though? Or is it the convention not to do this for function arguments.

john-romein:
I do not think that the wrappers need to enforce the use of anything newer than C++11,  The application itself could of course require a newer version.

include/cu.hpp
john-romein: I do not know what the general convention is, but in my opinion, the function prototype is a contract between the user and the implementation of the function.  For call-by-reference parameters, there is an obvious difference between const and nonconst arguments, where the implementation promises not to alter the parameter, but for call-by-value parameters, the const is an implementation detail that the user of the function should not be bothered with.  But this is just my opinion ...

include/cu.hpp
```diff
@@ -415,6 +419,12 @@ class Stream : public Wrapper<CUstream> {
 
   Stream(CUstream stream) : Wrapper<CUstream>(stream) {}
 
+  void memcpyHtoHAsync(void *dstPtr, const void *srcPtr, size_t size) {
```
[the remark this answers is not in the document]
csbnw: `memcpyHtoDAsync` and `memcpyDtoHAsync` also use `void*` for the host memory arguments, so I used the same for this method. Having the arguments type as template would also require a cast inside this function.

include/cu.hpp
csbnw: Same reason as above, the other functions also use `CUdeviceptr`.

include/cu.hpp
```diff
@@ -415,6 +419,12 @@ class Stream : public Wrapper<CUstream> {
...
+    checkCudaCall(cuMemcpyAsync(reinterpret_cast<CUdeviceptr>(dstPtr),
```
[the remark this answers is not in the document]
csbnw: I don't think it would be possible, because this function is intended to copy any host memory, whether is it allocated using `malloc`, or using `cu::HostMemory`. It is just that `cuMemcpyAsyc` requires the `CUdeviceptr`.

csbnw:
> I do not think that the wrappers need to enforce the use of anything newer than C++11, The application itself could of course require a newer version.

So the solution would be to disable the check in Codacy, as we can not work around it in C++11.

bouweandela:
Let's tackle the configuration of the linters in a separate pull request, we will need to decide on a style and settings that fit the entire library.

Issue #148, opened 2022-04-07
Add tests


bouweandela:
Ideas for tests:

Unit tests
- [ ] allocate some memory and copy to GPU and back
- [ ] compute fft with cufft and verify with fftw
- [ ] ...

Integration tests
- [ ] run and check vector add examples

Use CTest to run tests?

PR #169, opened 2022-07-15
Add a unit test and a framework for creating unit tests

bouweandela:
**Description**

Add some unit tests and the test framework [Catch2](https://github.com/catchorg/Catch2). I chose the Catch2 framework because it is the second most popular testing framework (after googletest), but it seems simpler.

**Related issues**:

- #148 

To try it out, run
```bash
cd $(mktemp -d -p . build-XXXX) && cmake -S .. -DBUILD_TESTING=ON && make && make test && cd -
```
To generate the coverage report `build/coverage.html`, run
```
gcovr
```
afterwards.

This pull request also includes the changes that I proposed separately in #170, so it might be easiest to review and merge that one first.

To finish the work on fixing the linter complaints/rules in #158, it would be really helpful to have some tests set up, so I can add more tests if I change something to check that I didn't break anything.


john-romein:
Maybe better to do
(cd $(mktemp -d -p . build-XXXX) && cmake -S .. -DBUILD_TESTING=ON && make && make test)
so that one stays in the current directory even if cmake or make fails.
Also: is it necessary to do "make && make test", instead of just a simple "make test"?  The test target of the Makefile could/should test the dependency on building all stuff.

CMakeLists.txt
```diff
@@ -1,5 +1,13 @@
 cmake_minimum_required(VERSION 3.17 FATAL_ERROR)
 
+# Find out if cudawrappers if used as a dependency. Build tests by default only
+# if not used as a dependency.
+if(NOT DEFINED PROJECT_NAME AND BUILD_TESTING)
+  set(CUDAWRAPPERS_TESTING_DEFAULT True)
+else()
+  set(CUDAWRAPPERS_TESTING_DEFAULT False)
+endif()
+
```
csbnw, line 10: Shouldn't every `CMakeLists` have a `project_name` directive? In other words, I don't think that we should test `NOT DEFINED PROJECT_NAME` here. Moreover, having the `CUDAWRAPPERS_` prefix to hide the variables doesn't seem like the cleanest solution either. How about using [CMake packaging](https://voxelated.github.io/CmakePackaging.html) with a `CudawrappersConfig.cmake` file?

CMakeLists.txt
```diff
@@ -9,24 +17,20 @@ project(
 )
 
 set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED True)
```
csbnw, line 20: Isn't this already implied by `CMAKE_CXX_STANDARD 11`?

CMakeLists.txt
```diff
@@ -9,24 +17,20 @@ project(
...
 set(CMAKE_CXX_EXTENSIONS False)
 set(CMAKE_BUILD_TYPE Release)
 option(BUILD_SHARED_LIBS "Create shared libraries" True)
-option(BUILD_TESTING "Built tests" False)
-include(CTest)
+option(CUDAWRAPPERS_BUILD_TESTING "Build cudawrappers tests"
+       ${CUDAWRAPPERS_TESTING_DEFAULT}
+)
+option(CUDAWRAPPERS_BUILD_EXAMPLES "Build cudawrappers examples" False)
```
csbnw, line 27: See also above, I don't think we should have the `CUDAWRAPPERS_` prefixes.

README.dev.md
```diff
@@ -96,6 +96,13 @@ To run the tests locally, enter the `build` directory and run:
 make test
 ```
 
+After running the tests, a coverage report `build/coverage.html` can be
+generated with [gcovr](https://gcovr.com/en/stable/installation.html) by running:
```
csbnw, line 100: Link to the main page of gcover?
```suggestion
generated with [gcovr](https://gcovr.com/en/stable/index.html) by running:
```

tests/CMakeLists.txt
```diff
@@ -1,2 +1,32 @@
-# vector_add
-add_subdirectory(vector_add)
+include(FetchContent)
+
+FetchContent_Declare(
+  Catch2
+  GIT_REPOSITORY https://github.com/catchorg/Catch2.git
+  GIT_TAG v2.13.9
```
csbnw, line 6: Is there a reason to use this specific (old) version?
The latest version is [v3.3.2](https://github.com/catchorg/Catch2/releases/tag/v3.3.2).

tests/test_cu.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+TEST_CASE("Test cu::Device", "[device]") {
+  cu::init();
+  cu::Device device(0);
+  cu::Context context(CU_CTX_SCHED_BLOCKING_SYNC, device);
+
+  SECTION("Test Device.getName") {
+    const std::string name = device.getName();
+    std::cout << "Device name: " << name << std::endl;
+    REQUIRE(name.size() > 0);
```
csbnw, line 18: In case only a single thing is tested, it is more appropriate to use `CHECK` rather than `REQUIRE`, see also [here](https://github.com/catchorg/Catch2/blob/devel/docs/assertions.md#natural-expressions).

tests/test_cu.cpp
```diff
@@ -0,0 +1,65 @@
...
+    const size_t size = sizeof(src);
+
+    cu::DeviceMemory mem(size);
+
+    cu::Stream stream;
+    stream.memcpyHtoDAsync(mem, src.data(), size);
+    stream.memcpyDtoHAsync(tgt.data(), mem, size);
+    stream.synchronize();
+
+    REQUIRE(src == tgt);
```
csbnw, line 40:
```suggestion
    CHECK(src == tgt);
```

tests/test_cu.cpp
```diff
@@ -0,0 +1,65 @@
...
+      tgt_ptr[i] = 0;
+    }
+    cu::DeviceMemory mem(size);
+
+    cu::Stream stream;
+    stream.memcpyHtoDAsync(mem, src, size);
+    stream.memcpyDtoHAsync(tgt, mem, size);
+    stream.synchronize();
+
+    REQUIRE(!static_cast<bool>(memcmp(src, tgt, size)));
```
csbnw, line 63:
```suggestion
    CHECK(!static_cast<bool>(memcmp(src, tgt, size)));
```

tests/vector_add/README.md
csbnw, line 1: This empty file should either be filled with some useful documentation, or removed.

tests/CMakeLists.txt
bouweandela: Yes, it's the last version that supports C++11. We can use the newer version if we drop support for C++11.

CMakeLists.txt
bouweandela: > Shouldn't every CMakeLists have a project_name directive? 

Yes, `PROJECT_NAME` is defined once `project()` has been called. For this file, that is done on line 11 below, therefore this file must have been included from another project if `PROJECT_NAME` is already defined here.

> Moreover, having the CUDAWRAPPERS_ prefix to hide the variables doesn't seem like the cleanest solution either. How about using [CMake packaging](https://voxelated.github.io/CmakePackaging.html) with a CudawrappersConfig.cmake file?

We are using CMake packaging, see: https://github.com/nlesc-recruit/cudawrappers/blob/c9dafd98aed017919289451ec2d29008d4d68503/CMakeLists.txt#L97-L113
but how do I make use of CMake packaging to replace these variables?

CMakeLists.txt
bouweandela: According to the CMake documentation [here](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD_REQUIRED.html#prop_tgt:CXX_STANDARD_REQUIRED), `CMAKE_CXX_STANDARD 11` is treated as a sort of nice to have recommendation instead of a requirement unless you also set this variable to `True`.

tests/vector_add/README.md
bouweandela: I can't see it on my computer when I look at this branch of the git repository, so I think I removed it in this pull request and that's why it is listed here. Do you see it if you check out this branch?

README.dev.md
bouweandela: Changed in [1ce83a9](https://github.com/nlesc-recruit/cudawrappers/pull/169/commits/1ce83a981228b18a6a7b82dbf2180de755482d56)

tests/CMakeLists.txt
bouweandela: As far as I'm aware C++ is not forward-compatible, so using the latest version of Catch2 would mean that we cannot run the tests in C++11 so you would have no idea if your code actually works in C++11, making it hard to claim that that version of the language is supported.

tests/test_cu.cpp
bouweandela: Done in [1ce83a9](https://github.com/nlesc-recruit/cudawrappers/pull/169/commits/1ce83a981228b18a6a7b82dbf2180de755482d56)

tests/test_cu.cpp
bouweandela: Done in [1ce83a9](https://github.com/nlesc-recruit/cudawrappers/pull/169/commits/1ce83a981228b18a6a7b82dbf2180de755482d56)

CMakeLists.txt
csbnw: As we discussed, prefixing with `CUDAWRAPPERS_` is the most straightforward way of getting the desired functionallity.

CMakeLists.txt
csbnw: You are right, we need both in this case.

tests/CMakeLists.txt
csbnw: I have created [this](https://github.com/nlesc-recruit/cudawrappers/issues/186) feature request for upgrading to a newer version of Catch2.

tests/vector_add/README.md
csbnw: It is indeed not present when cloning the branch.

csbnw (approved):

PR #170, opened 2022-07-28
Improve includes

### LINKED ISSUE #166: Include as <cudawrappers/cu.hpp> instead of <cu.hpp>
It would be nicer to be able to include as <cudawrappers/cu.hpp> instead of <cu.hpp>.

bouweandela:
**Description**

- Put the headers in a `cudawrappers` directory, so they can be included with e.g. `#include <cudawrappers/cu.hpp>`
- Use [include-what-you-use](https://github.com/include-what-you-use/include-what-you-use#using-with-cmake) to get the includes right
- Sort includes according to [Google style guide](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes)

**Related issues**:

Closes #166


.clang-format
```diff
@@ -81,8 +81,8 @@ IfMacros:
   - KJ_IF_MAYBE
 IncludeBlocks:   Regroup
 IncludeCategories:
-  - Regex:           '^<ext/.*\.h>'
-    Priority:        2
+  - Regex:           '^<catch2/catch.hpp|cuda.h|cuda_runtime_api.h|nvrtc.h|cudawrappers/.*\.hpp>'
+    Priority:        3
```
csbnw, line 85: Why is this change needed? Shouldn't the `.clang-format` be as generic as possible, without `cudawrappers` specifics, such that the same file can be reused for multiple projects to have consistent formatting?

.clang-format
```diff
@@ -94,11 +94,11 @@ IncludeCategories:
     SortPriority:    0
     CaseSensitive:   false
   - Regex:           '.*'
-    Priority:        3
```
csbnw, line 97: Same as above: I don't think these lines should be changed.

.clang-format
csbnw: I like having the includes property sorted! However, can't the `.clang-format` somehow not be more generic to automatically consider all header files not in the source tree as third-party?

bouweandela: It looks like clang-format uses just these regular expressions for determining which category an include is. Without these modifications, standard library includes and third-party includes will become a single group because clang-format does not know the distinction. Alternatively, we could make a list of all standard library includes, anything that is not in it will be third-party. That would have the advantage that it is generic across projects. Do you think that would be better?

csbnw: Creating and maintaining an exhaustive list of standard libraries is probably more of a hassle than your current approach.

Now that we have discussed the possibilities, I feel like this thread can be closed.

csbnw:
Regarding [fde509c](https://github.com/nlesc-recruit/cudawrappers/pull/170/commits/fde509c02a3043beb6059af75c7be7e3dabd50fb), consider rebasing onto the target branch rather than merging. This keeps the history cleaner.

bouweandela:
Would it be OK to merge with the 'Squash and merge' button instead of rebasing?

bouweandela:
@csbnw @john-romein Any further comments on this pull request? Otherwise, I think it is ready to be merged.

csbnw:
The pre-commit check fail. Apart from that, it looks good to me and I think it is ready to be merged.

PR #176, opened 2023-05-22
pre-commit CI and linter updates

### LINKED ISSUE #172: Pre-commit GitHub Action checks are failing
The checks run by the GitHub Action [pre-commit](https://github.com/nlesc-recruit/cudawrappers/blob/main/.github/workflows/pre-commit.yml) are failing, output from a [recent run](https://github.com/nlesc-recruit/cudawrappers/actions/runs/4414176403/jobs/7735550369):
```
clang-tidy...............................................................Failed
- hook id: clang-tidy
- exit code: 1

/home/runner/work/cudawrappers/cudawrappers/tests/vector_add/vector_add.cpp:11:11: warning: 1024 is a magic number; consider replacing it with a named constant [cppcoreguidelines-avoid-magic-numbers]
  int N = 1024;
          ^
/home/runner/work/cudawrappers/cudawrappers/tests/vector_add/vector_add.cpp:12:21: warning: implicit conversion changes signedness: 'int' to 'unsigned long' [clang-diagnostic-sign-conversion]
  size_t bytesize = N * sizeof(float);
                    ^ ~
/home/runner/work/cudawrappers/cudawrappers/tests/vector_add/vector_add.cpp:21:3: warning: use auto when initializing with a cast to avoid duplicating the type name [modernize-use-auto]
  float *a = (float *)h_a;
  ^~~~~
  auto
/home/runner/work/cudawrappers/cudawrappers/tests/vector_add/vector_add.cpp:21:14: warning: use of old-style cast [clang-diagnostic-old-style-cast]
  float *a = (float *)h_a;
             ^        ~~~
/home/runner/work/cudawrappers/cudawrappers/tests/vector_add/vector_add.cpp:22:3: warning: use auto when initializing with a cast to avoid duplicating the type name [modernize-use-auto]
  float *b = (float *)h_b;
  ^~~~~
  auto
/home/runner/work/cudawrappers/cudawrappers/tests/vector_add/vector_add.cpp:22:14: warning: use of old-style cast [clang-diagnostic-old-style-cast]
  float *b = (float *)h_b;
             ^        ~~~
/home/runner/work/cudawrappers/cudawrappers/tests/vector_add/vector_add.cpp:23:3: warning: use auto when initializing with a cast to avoid duplicating the type name [modernize-use-auto]
  float *c = (float *)h_c;
  ^~~~~
  auto
/home/runner/work/cudawrappers/cudawrappers/tests/vector_add/vector_add.cpp:23:14: warning: use of old-style cast [clang-diagnostic-old-style-cast]
  float *c = (float *)h_c;
             ^        ~~~
/home/runner/work/cudawrappers/cudawrappers/tests/vector_add/vector_add.cpp:25:5: warning: do not use pointer arithmetic [cppcoreguidelines-pro-bounds-pointer-arithmetic]
    a[i] = 1.0;
    ^
/home/runner/work/cudawrappers/cudawrappers/tests/vector_add/vector_add.cpp:26:5: warning: do not use pointer arithmetic [cppcoreguidelines-pro-bounds-pointer-arithmetic]
    b[i] = 1.0;
    ^
/home/runner/work/cudawrappers/cudawrappers/tests/vector_add/vector_add.cpp:27:5: warning: do not use pointer arithmetic [cppcoreguidelines-pro-bounds-pointer-arithmetic]
    c[i] = 0.0;
    ^
/home/runner/work/cudawrappers/cudawrappers/tests/vector_add/vector_add.cpp:41:10: error: 'vector_add_kernel.cu' file not found [clang-diagnostic-error]
#include "vector_add_kernel.cu"
         ^~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/cudawrappers/cudawrappers/tests/vector_add/vector_add.cpp:56:29: warning: variable 'parameters' is not initialized [cppcoreguidelines-init-variables]
  std::vector<const void *> parameters = {d_c.parameter(), d_a.parameter(),
                            ^
                                       = 0
/home/runner/work/cudawrappers/cudawrappers/tests/vector_add/vector_add.cpp:65:30: warning: do not use pointer arithmetic [cppcoreguidelines-pro-bounds-pointer-arithmetic]
  std::cout << "hurray! " << c[0] << " \n";
                             ^
4532 warnings and 1 error generated.
Error while processing /home/runner/work/cudawrappers/cudawrappers/tests/vector_add/vector_add.cpp.
Suppressed 4519 warnings (4519 in non-user code).
Use -header-filter=.* to display errors from all non-system headers. Use -system-headers to display errors from system headers as well.
Found compiler error(s).

cppcheck.................................................................Failed
- hook id: cppcheck
- exit code: 1

src/nvtx.cpp:6:3: performance: Variable '_attributes' is assigned in constructor body. Consider performing initialization in initialization list. [useInitializationList]
  _attributes = {0};
  ^
include/nvtx.hpp:12:3: style: Class 'Marker' has a constructor with 1 argument that is not explicit. [noExplicitConstructor]
  Marker(const char* message, unsigned color = 0xff00ff00);
  ^
src/nvtx.cpp:6:3: performance: Variable '_attributes' is assigned in constructor body. Consider performing initialization in initialization list. [useInitializationList]
  _attributes = {0};
  ^
include/nvtx.hpp:12:3: style: Class 'Marker' has a constructor with 1 argument that is not explicit. [noExplicitConstructor]
  Marker(const char* message, unsigned color = 0xff00ff00);
  ^
```

loostrum:
Pinning compiler and linter versions in environment.yml and using this conda environment in the pre-commit CI.
Will also fix linter errors. Then this PR closes #172


loostrum:
Interestingly some errors that the CI is catching (see https://github.com/nlesc-recruit/cudawrappers/actions/runs/5046918293/jobs/9053071511#step:10:65)
are not caught on my machine. The pointer arithmetic errors did show up at first, but disappeared later after fixing some other things. `pre-commit run -a` does not catch them. `clang-tidy -p build/ tests/vector_add/vector_add.cpp` does catch them. I'm not sure what the exact arguments to clang-tidy are as used by pre-commit.

loostrum:
@bouweandela I fixed most of the linter issues, but I'm stuck on a missing include of the vector_add kernel in the tests. pre-commit doesn't fail on this locally so I'm not sure whats wrong. Could you explain how the cmake stuff that handles the include of cuda code works? If I change  `#include "vector_add_kernel.cu"` to ` #include "kernels/vector_add_kernel.cu"` the include works, but compilation fails because the code isn't actually included as a string (I think).
```[ 87%] Building CXX object tests/vector_add/CMakeFiles/vector_add.dir/vector_add.cpp.o
In file included from /home/oostrum/recruit/cudawrappers/tests/vector_add/vector_add.cpp:43:
/home/oostrum/recruit/cudawrappers/tests/vector_add/kernels/vector_add_kernel.cu: In function 'void vector_add()':
/home/oostrum/recruit/cudawrappers/tests/vector_add/kernels/vector_add_kernel.cu:1:1: error: expected primary-expression before 'extern'
    1 | extern "C" __global__ void vector_add(float *c, float *a, float *b, int n) {
      | ^~~~~~
```

loostrum:
building with -DBUILD_TESTING=True seems to have fixed it, all green now!

loostrum:
@csbnw there are mostly style changes in this PR, but also a few constructors were made `explicit` so please check if this would break any existing code. I don't think so, but better safe than sorry.

bouweandela:
> Edit: I just found [this](https://stackoverflow.com/questions/410980/include-a-text-file-in-a-c-program-as-a-char) which looks like what you did?

Yes, I think the implementation here is based on the answer by 'Martin R'

environment.yml
```diff
@@ -2,21 +2,21 @@ name: recruit
...
+  - cudatoolkit-dev==10.1.243
+  - cxx-compiler==1.5.2
   - doxygen
   - include-what-you-use
   - make
   - pip
-  - pre-commit
-  - python
-  - yamllint
+  - python==3.11.3
```
bouweandela, line 16: I would only pin linters and automatic formatters (and probably clang because it is used by some of them), but not the packages that should not affect linting/formatting. So I would suggest to unpin `cmake, python, pre-commit` and possibly also `c-compiler, cxx-compiler, cudatoolkit-dev`. Does that make sense?

cmake/linter-tools.cmake
```diff
@@ -1,30 +1,60 @@
...
+#
 # Ref:
 # https://stackoverflow.com/questions/32280717/cmake-clang-tidy-or-other-script-as-custom-target
 
 file(GLOB_RECURSE ALL_SOURCE_FILES src/*.cpp include/*.hpp)
 file(GLOB_RECURSE ALL_CMAKE_LISTS CMakeLists.txt *.cmake)
 
-add_custom_target(clang-format COMMAND clang-format -i ${ALL_SOURCE_FILES})
+add_custom_target(
+  clang-format
```
bouweandela, line 13: Do we even need all these custom targets if we already have pre-commit? This looks like duplicate functionality

bouweandela:
> I'm not sure what the exact arguments to clang-tidy are as used by pre-commit.

What it comes down to is that our pre-commit configuration here:
https://github.com/nlesc-recruit/cudawrappers/blob/41f091c100bffbb88df05bd6f749f2e5b06e8ae5/.pre-commit-config.yaml#L6-L11
defines how we call the hook defined here:
https://github.com/pocc/pre-commit-hooks/blob/336fdd7c3cab698ead0b1c95157b9e74d3906b62/.pre-commit-hooks.yaml#L12-L17
which calls the executable defined here:
https://github.com/pocc/pre-commit-hooks/blob/336fdd7c3cab698ead0b1c95157b9e74d3906b62/setup.cfg#L28
which is this Python script:
https://github.com/pocc/pre-commit-hooks/blob/master/hooks/clang_tidy.py

I suspect the pre-commit hooks will only work correctly if you call them from the root of this repository and you call the build directory `build` because the `-p` flag is a relative path to the `build` directory. Maybe you did something slightly different on your own machine?

loostrum:
Thanks, I think something related to the build directory happened indeed because the CI wasn't building the test code but I was doing that locally. So the compilation database was different. In any case it's now passing both locally and on the CI so it should be fixed.

environment.yml
loostrum: Good point, I'll unpin as you suggest.

cmake/linter-tools.cmake
loostrum: I don't think we do actually, let's remove them.

.pre-commit-config.yaml
```diff
@@ -16,14 +16,9 @@ repos:
     hooks:
       - id: cmake-format
       - id: cmake-lint
-  - repo: https://github.com/citation-file-format/cff-converter-python
```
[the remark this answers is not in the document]
loostrum: This one is duplicated by the cffconvert workflow we also have running on the CI. I thought it would be nicer to have that as a separate CI on Github than to have it as part of pre-commit, because it's not related to the code unlike everything else in pre-commit

.github/workflows/pre-commit.yml
```diff
@@ -22,47 +22,52 @@ jobs:
...
+
+      - uses: actions/cache@v2
+        with:
+          path: /usr/share/miniconda/envs/recruit
+          key: conda-${{ hashFiles('environment.yml') }}
+        id: cache
+
+      - name: Update mamba environment if cache is outdated
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: mamba env update -n recruit -f environment.yml
```
bouweandela, line 62: I would maybe delete the environment and create a new one from scratch instead of trying to update, as this tends to be slow and may fail.

loostrum: I'll double-check what's happening in this step - if the cache misses there is no environment so it should be equivalent to creating a new env anyway. Better to do that explicitly then.

loostrum: My previous comment isn't actually correct: there _is_ already an environment (created in the mamba setup step), but it's empty. I tried updating an empty env vs creating a new one and the update is actually a bit faster so I suggest leaving this as-is. Assuming that updating an empty env won't fail that is. This mamba issue may be why update is actually faster: https://github.com/mamba-org/mamba/issues/633

bouweandela:
Looks good to me now! Could you please address the merge conflict?

loostrum:
Thanks, done!

Issue #194, opened 2023-07-14
Clang-tidy incorrectly warns about system headers, in CI only


loostrum:
**Describe the bug**
Clang-tidy gives errors about headers that are not part of cudawrappers. See e.g. [this CI run](https://github.com/nlesc-recruit/cudawrappers/actions/runs/5553197146/jobs/10141454485). Locally, this issue is not present.
I tried adding `HeaderFilterRegex = "*cudawrappers*"` to force clang-tidy to only scan the correct headers, but this made no difference. For now I'm suggesting to disable clang-tidy in CI so we can continue working on the code. This means users _should_ run it locally for now. We can then investigate the clang-tidy issue separately from features PRs.

**To Reproduce**
Push something with a style issue to github, clang-tidy will complain about this _and_ system header issues.

**Expected behavior**
Clang-tidy should
1) give consistent results locally and in the CI
2) not complain about system headers

loostrum:
We are now working on cufft without clang tidy in the CI (see cufft PR: https://github.com/nlesc-recruit/cudawrappers/pull/151)
When clang-tidy is re-enabled, we should make sure to check the cufft files

bouweandela:
The warnings look like they could be compiler errors. Are we sure we have all the required software installed in CI to compile? Related to #197

PR #195, opened 2023-07-14
Temporarily disable clang-tidy

loostrum:
**Description**

Temporarily disable clang-tidy, until we fix the spurious errors it gives about system headers.

**Related issues**:
#194


csbnw:
I would suggest to either:
1) add a comment to explain why the code is commented,
2) or remove it completely
The second option has my preference, as we can always find the deleted code back in the history.

loostrum:
Sounds good, I've removed the code completely. We can easily find it back in this PR.

csbnw (approved):

PR #199, opened 2023-07-28
Suggestion: improve CI runtime

### LINKED ISSUE #189: Check use of conda cache in CI
The idea was that the conda environment is cached and re-used in the CI, but that doesn't seem to happen. Every run of the CI rebuilds the conda environment. This should either be fixed, or we might remove the conda caching altogether. Another option to speed up runs of the CI is to build a Docker container containing the necessary software.

loostrum:
**Description**

In an attempt to fix #189 I made an alternative pre-commit CI workflow that doesn't use conda. The runtime improved from ~7 minutes to <2 minutes.

This PR is one way of improving the CI runtime. Without conda, we depend on the tools available in the Github runner and apt/pip to install the missing ones. Most tools we need are already provided by the runner. I'm using apt/pip to install the missing linter/formatter tools, and a specific github action to install cuda (only the required parts of cuda, not everything as that takes a very long time).

The downside of this approach vs conda is that we are less flexible in the versions of some packages. The Github runners typically provide the latest 3 major versions of compilers etc, but for e.g. cppcheck we depend on apt and whatever version is available there. This is not necessarily a problem, but good to be aware of. The CI now prints the versions of all used tools as well, so one can more easily check if the local versions match. I've updated the versions in `environment.yml` to match the CI.

The current approach does make it easy to test with e.g. several CUDA versions using the Github actions matrix feature, but this is not implemented yet.

**Related issues**:
#189 describes a problem with the conda cache. This is fixed by removing conda completely. The conda environment build step typically took about 5 minutes. Now the major factor in the CI is installing cuda, which takes ~1 minute. `cmake` warns about not finding cublas, but everything else passes so I don't think we need it. Adding it increases build times by >50% so I've left it out for now.

As part of testing I forced the CI to run everything with `pre-commit run -a` instead of the current command, and everything passed.

- Verify that the CI passes
- Verify that the versions listed in the CI output and in environment.yml are the same

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -55,7 +55,7 @@ jobs:
         with:
           cuda: ${{ env.cuda-version }}
           method: "network"
-          sub-packages: '["nvcc"]'
+          sub-packages: '["nvcc", "nvrtc", "nvtx", "cufft"]'
           linux-local-args: '["--toolkit"]'
 
       - name: Generate compilation database
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -12,9 +12,11 @@ jobs:
         # set tools versions to latest supported by the CI runner,
         # ubuntu 22.04 at the time of writing
         # clang defaults to v14.0.0
-        gcc-version: 12  # = 12.1.0
+        gcc-version: "12"  # = 12.1.0
         # cuda version to install with separate action
         cuda-version: "12.2.0"
+        # version for cmake-lint and cmake-format
+        cmakelang-version: "0.6.13"
     steps:
 
       - name: Show git version
@@ -61,7 +63,7 @@ jobs:
 
       - name: Install pre-commit and related tools
         run: |
-          pip install pre-commit cmake-lint cmake-format
+          pip install pre-commit cmakelang==${{ env.cmakelang-version}}
           sudo apt install cppcheck -y
 
       - name: Show tool versions
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -59,9 +59,9 @@ jobs:
           non-cuda-sub-packages: '["libcufft-dev", "libcublas-dev"]'
           linux-local-args: '["--toolkit"]'
 
-      - name: Install pre-commit and cppcheck
+      - name: Install pre-commit and related tools
         run: |
-          pip install pre-commit
+          pip install pre-commit cmake-lint cmake-format
           sudo apt install cppcheck -y
 
       - name: Show tool versions
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,5 +1,4 @@
 name: pre-commit
-# conda cache setup based on https://github.com/epassaro/cache-conda-envs
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


laurasootes:
I tried to run the CI on DAS6, but I did not manage to get it to pass
My steps were:
- load spack
- load cuda
- load cmake
- load gcc
- load python
- load py-pip 
- pip install pre-commit

run pre-commit run -a
clang-format, cmake-format, and cmake-lint pass, but cppcheck does not. It gives 
`FileNotFoundError: [Errno 2] No such file or directory: 'cppcheck'`
Did I miss anything? ccpcheck should not be installed as well, as it should be installed during the run, right?

on a second appempt, cmake-format and make-lint fail as well.
`ModuleNotFoundError: No module named 'six'`

csbnw:
> on a second attempt, cmake-format and cmake-lint fail as well. raised_eyebrow `ModuleNotFoundError: No module named 'six'`
On DAS-6, many common Python packages can be loaded by running `module load py-<name-of-package>`. So could you try to see if `module load py-six` works?

laurasootes:
Ah yes, I overlooked that. It solves the `ModuleNotFoundError: No module named 'six'`, thanks!

That only leaves me with the cppcheck failing. Does it work on your side @csbnw?

csbnw:
It works for me. After loading `spack/13.1.0`, you will have `gcc/13.1.0` loaded automatically. The only thing I needed to load additionally was `cppcheck/2.9`. Moreover, `pre-commit` seems to assume a `build` directory inside the source tree. I fixed this by adding a symbolic link. Afterwards, it runs fine:
```
$ pre-commit run -a
clang-format.............................................................Passed
cppcheck.................................................................Passed
cmake-format.............................................................Passed
cmake-lint...............................................................Passed
```

laurasootes:
after following your steps:
`module load spack/13.1.0 py-pre-commit/2.17.0 cppcheck/2.9`

I only get the error:
`cppcheck: error: failed to open project 'build/compile_commands.json'. The file does not exist.`
I created the build folder, but it then still does not have the compile_commands.json, hence the error. Do you know where I should get that file from?

I am still confused as to why the cppcheck module has be loaded, if part of the pre-commit.yml runs an install for it?

```     
- name: Install pre-commit and related tools
  run: |
    pip install pre-commit cmakelang==${{ env.cmakelang-version}}
    sudo apt install cppcheck -y
```

csbnw:
> I only get the error:
> `cppcheck: error: failed to open project 'build/compile_commands.json'. The file does not exist.`
> I created the build folder, but it then still does not have the compile_commands.json, hence the error. Do you know where I should get that file from?

You need to run `cmake` (and possibly also `make`?) in that build directory prior to running `pre-commit`. That way, your build directory will have the required `compile_commands.json` file.

csbnw:
> I am still confused as to why the cppcheck module has be loaded, if part of the pre-commit.yml runs an install for it?
> 
> ```
> - name: Install pre-commit and related tools
>   run: |
>     pip install pre-commit cmakelang==${{ env.cmakelang-version}}
>     sudo apt install cppcheck -y
> ```

This code only runs on the CI. When running `pre-commit` manually, you don't install any packages using the package manager but load the software using environment modules instead.

csbnw (approved):

Issue #133, opened 2022-03-07
Links in the online documentation do not work


fdiblen:
README.md has links to README.dev.md and CONTRIBUTING.md. When hosting the documenatiton on readthedocs, these links do not work.

csbnw:
See https://github.com/nlesc-recruit/cudawrappers/pull/203.
Unfortunately, the problem still persists.

csbnw:
We looked into this again, but still can't seem to fix it, as the examples shown in the documentation simply don't work. Therefore, we will postpone fixing this issue for some future release (not 0.5.0).

laurasootes:
@csbnw 

It seems to work when using the full weblinks to the files (see [branch](https://github.com/nlesc-recruit/cudawrappers/tree/readthedocs_links)). I recreated the readthedocs page from a fork, and there the links do work:
[fork rtd](https://cudawrapperss.readthedocs.io/en/latest/#)

Or is it undesirable to do it this way?

csbnw:
Using the full weblinks was indeed one of the things that was considered before. It is not a very elegant solution, arguably better than having links that do not function at all.

PR #203, opened 2023-09-13
!133 Update read the docs

csbnw:
**Description**

On [readthedocs](https://cudawrappers.readthedocs.io/en/latest/), the links to `CONTRIBUTING` and `CHANGELOG.md` are not clickable. These have been changed to be links to this repository. This is a workaround, since readthedocs has the option to view different versions of the documentation which now all link to files on the main branch.
As an alternative, we could point to these files at a specific tag, but this comes with the downside of always having to update the links upon a new release.

Additionally, the names of the alternatives to cudawrappers are now clickable.

**Related issues**:

- https://github.com/nlesc-recruit/cudawrappers/issues/133

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,9 +78,7 @@ This section aims to provide an overview of projects that use this repo's librar
 
 This section provides an overview of similar tools in this space, and how they are different.
 
-### cuda-api-wrappers
-
-url: https://github.com/eyalroz/cuda-api-wrappers
+### [cuda-api-wrappers](https://github.com/eyalroz/cuda-api-wrappers)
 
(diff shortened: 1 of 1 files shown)


csbnw:
@matmanc and I tested whether the links works correctly by building the documentation locally and opening the webpage. It doesn't work for some reason, not even after upgrading to newer sphinx etc. packages. According to the documentation, we should now be using the correct syntax, so readthedocs will work after merging anyway. Let's give it a try? If it doesn't work, the markdown is at least a bit better than it used to be.

matmanc (approved):

Issue #187, opened 2023-06-23
Fix Device()


laurasootes:
https://github.com/nlesc-recruit/cudawrappers/blob/db1f3ff9e6e9dbffe51cb0d648f3e0a4c0023280/include/cudawrappers/cu.hpp#L190C1-L190C48

Fix Device in  cu.hpp -> Device getDevice()

matmanc:
Fixed in https://github.com/nlesc-recruit/cudawrappers/pull/204.

PR #204, opened 2023-09-13
!167 Remove deprecated Context::getDevice

csbnw:
**Description**

There was as `FIXME` at the `Context::getDevice` function, related to the destructor.  It appears that the function isn't needed at all. One should instantiate `cu::Device` instances by calling the constructor of `Device` and pass the `Context` as an argument. Therefore, `Context::getDevice` is deprecated and now removed.

**Related issues**:

- https://github.com/nlesc-recruit/cudawrappers/issues/187

### DIFF
--- a/include/cudawrappers/cu.hpp
+++ b/include/cudawrappers/cu.hpp
@@ -188,12 +188,6 @@ class Context : public Wrapper<CUcontext> {
     checkCudaCall(cuCtxSetSharedMemConfig(config));
   }
 
-  static Device getDevice() {
-    CUdevice device{};
-    checkCudaCall(cuCtxGetDevice(&device));
-    return Device(device);  // FIXME: ~Device()
-  }
-
   static size_t getLimit(CUlimit limit) {
     size_t value{};
     checkCudaCall(cuCtxGetLimit(&value, limit));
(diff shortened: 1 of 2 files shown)


include/cudawrappers/cu.hpp
```diff
@@ -188,12 +188,6 @@ class Context : public Wrapper<CUcontext> {
     checkCudaCall(cuCtxSetSharedMemConfig(config));
   }
 
-  static Device getDevice() {
```
matmanc, line 191: wouldnt be better to add a deprecation warning?
Maybe someone is using it

csbnw: Some time ago, we agreed that the CHANGELOG is the place to explicitly mention API breaking changes. Users typically use tagged versions, so they should be able to be aware of these changes. Moreover, since cudawrappers is currently still evolving (and changing) quite rapidly, it is very likely that some refactoring is needed upon upgrading to a newer version anyway. Therefore, I would prefer to simply remove the function instead of keeping 'legacy' code that no one really uses and/or knows/cares about.

matmanc: would it be an idea to add the suggested refactoring in the CHANGELOG?

csbnw: The CHANGELOG is already updated to reflect this change, see [here](https://github.com/nlesc-recruit/cudawrappers/pull/204/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed). Or do you mean someone else?

matmanc (approved):

PR #206, opened 2023-09-13
Update CMake for unit tests

csbnw:
**Description**

The tests executables `test_cu`, `test_nvrtc` and `test_cufft` were build, but running `ctest` or `make test` didn't detect them. This is now solved by changing the `CMakeLists`.

**Related issues**:

- https://github.com/nlesc-recruit/cudawrappers/issues/148

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -12,25 +12,23 @@ FetchContent_MakeAvailable(Catch2)
 list(APPEND CMAKE_MODULE_PATH ${catch2_SOURCE_DIR}/contrib)
 include(Catch)
 
-add_executable(test_cu test_cu.cpp)
-target_link_libraries(test_cu PUBLIC Catch2::Catch2 cudawrappers::cu)
-catch_discover_tests(test_cu)
-
-add_executable(test_nvrtc test_nvrtc.cpp)
-target_link_libraries(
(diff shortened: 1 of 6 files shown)


tests/CMakeLists.txt
```diff
@@ -12,23 +12,25 @@ FetchContent_MakeAvailable(Catch2)
...
 include(Catch)
 
-# Compile Catch2
-add_library(tests)
-target_sources(tests PRIVATE main.cpp)
-target_link_libraries(tests PUBLIC Catch2::Catch2)
-set(LINK_cu cudawrappers::cu)
-set(LINK_cufft cudawrappers::cu cudawrappers::cufft)
-set(LINK_nvrtc cudawrappers::nvrtc)
+add_executable(test_cu test_cu.cpp)
```
matmanc, line 15: it seams like there is a pattern to add each test maybe you could consider making it a function

csbnw: I considered this, especially since the previous code also used a foreach loop. Not every test however needs to be linked to the same `cudawrappers::*` libraries, and I thought that some repeated code was more comprehensive in this case.

matmanc (approved):

tests/CMakeLists.txt
csbnw: Some duplicated code could indeed be removed.

Issue #197, opened 2023-07-28
Compile code in CI


loostrum:
**Is your feature request related to a problem? Please describe.**
While we `cmake` as part of the `pre-commit` CI, we don't actually compile the code in any of the CI workflows. 

**Describe the solution you'd like**
I suggest to add a workflow to compile the code. A single OS/compiler combination seems enough, but we could consider adding a matrix with several, e.g. both g++ and clang, several cuda versions, etc.

PR #213, opened 2023-09-20
Add github workflow to build the code

csbnw:
**Description**

As proposed in #197, the code is now actually being build in the CI by adding a new workflow to do so.

**Related issues**:

- https://github.com/nlesc-recruit/cudawrappers/issues/197

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,10 +9,7 @@ jobs:
     name: run
     runs-on: ubuntu-latest
     env:
-        # set tools versions to latest supported by the CI runner,
-        # ubuntu 22.04 at the time of writing
         gcc-version: "12" 
-        # cuda version to install with separate action
         cuda-version: "12.2.0"
     steps:
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ jobs:
     env:
         # set tools versions to latest supported by the CI runner,
         # ubuntu 22.04 at the time of writing
-        gcc-version: "12"  # = 12.1.0
+        gcc-version: "12" 
         # cuda version to install with separate action
         cuda-version: "12.2.0"
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,52 @@
...
+  workflow_dispatch:
+
+jobs:
+  linting:
+    name: run
+    runs-on: ubuntu-latest
+    env:
+        # set tools versions to latest supported by the CI runner,
+        # ubuntu 22.04 at the time of writing
+        gcc-version: "12"  # = 12.1.0
```
matmanc, line 14: this is actually 12.3.0 at the time of running.

Maybe just remove the comment as it is misleading

matmanc (approved):
Looks good except fo the comment

.github/workflows/build.yml
csbnw, line 14:
```suggestion
        gcc-version: "12" 
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,52 @@
...
+  push:
+  workflow_dispatch:
+
+jobs:
+  linting:
+    name: run
+    runs-on: ubuntu-latest
+    env:
+        # set tools versions to latest supported by the CI runner,
+        # ubuntu 22.04 at the time of writing
```
csbnw, line 13: This comment is not really helpful and can be removed:
```suggestion
        # set tools versions to latest supported by the CI runner,
        # ubuntu 22.04 at the time of writing
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,52 @@
...
+        gcc-version: "12" 
+        # cuda version to install with separate action
```
csbnw, line 15: The code is fairly obvious, so the comment is redudant.
```suggestion
        # cuda version to install with separate action
```

.github/workflows/build.yml
csbnw: You are right. The comment is now removed. I also removed the other comments, as they are not very helpfull either.

PR #218, opened 2023-09-22
Add functions to asynchronously allocate and free device memory

csbnw:
**Description**

`Stream::memAllocAsync` and `Stream::memFreeAsync` are added. To test this new functionality, `Context::getFreeMemory` (and for completeness `Context::getTotalMemory`) have also been added so that a test can be performed to see if the free device memory before and after using the asynchronous allocation remains the same. The test is not performed on devices with integrated memory, as we have no control over other processes allocating/free-ing the memory.

### DIFF
--- a/tests/test_vector_add.cpp
+++ b/tests/test_vector_add.cpp
@@ -130,6 +130,39 @@ TEST_CASE("Vector add") {
     }
   }
 
+  SECTION("Run kernel with asynchronously allocated memory") {
+    cu::HostMemory h_a(bytesize);
+    cu::HostMemory h_b(bytesize);
+    cu::HostMemory h_c(bytesize);
+    std::vector<float> reference_c(N);
+
+    initialize_arrays(static_cast<float *>(h_a), static_cast<float *>(h_b),
+                      static_cast<float *>(h_c), reference_c.data(), N);
+
+    const size_t memory_free = context.getFreeMemory();
+
+    cu::DeviceMemory d_a = stream.memAllocAsync(bytesize);
+    cu::DeviceMemory d_b = stream.memAllocAsync(bytesize);
+    cu::DeviceMemory d_c = stream.memAllocAsync(bytesize);
+
+    stream.memcpyHtoDAsync(d_a, h_a, bytesize);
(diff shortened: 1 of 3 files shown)


john-romein (commented):
Is CHECK(memory_free == context.getFreeMemory()); necessarily true?  What, for example, if somebody else uses the same GPU at the same time?

csbnw:
> Is CHECK(memory_free == context.getFreeMemory()); necessarily true? What, for example, if somebody else uses the same GPU at the same time?

It will indeed fail when someone else is using the same GPU at the same time. But while running the test during development or in CI (we don't have this yet), a dedicated GPU will likely be used and this is not a problem.

john-romein:
Maybe better to leave it out then, as it looks like an assertion, and assertions should not fail because of external conditions.

csbnw:
> Maybe better to leave it out then, as it looks like an assertion, and assertions should not fail because of external conditions.

Ok, I will comment out `CHECK` and add a comment. Maybe once cudawrappers also supports NVML, we can try to find out whether the test is the only process using the GPU and enable the test accordingly.

tests/test_vector_add.cpp
```diff
@@ -129,4 +129,37 @@ TEST_CASE("Vector add") {
...
+    stream.launchKernel(function, 1, 1, 1, N, 1, 1, 0, parameters);
+    stream.memcpyDtoHAsync(h_c, d_c, bytesize);
+    stream.memFreeAsync(d_a);
+    stream.memFreeAsync(d_b);
+    stream.memFreeAsync(d_c);
+    stream.synchronize();
+
+    if (!device.getAttribute(CU_DEVICE_ATTRIBUTE_INTEGRATED)) {
+      CHECK(memory_free == context.getFreeMemory());
+    }
```
csbnw, line 161:
```suggestion
    if (!device.getAttribute(CU_DEVICE_ATTRIBUTE_INTEGRATED)) {
      // The following CHECK is disabled since we cannot make sure that the test is using the GPU exclusively
      // CHECK(memory_free == context.getFreeMemory());
    }
```

john-romein (approved):

Issue #198, opened 2023-07-28
Consider making cudawrappers header-only


loostrum:
**Is your feature request related to a problem? Please describe.**
The tensor-core correlator requires header-only cudawrappers. This avoids errors when using multiple architectures and having to manage libraries that are compiled for that specific architecture. Pinging @john-romein in case there are more relevant details to share regarding the correlator.

**Describe the solution you'd like**
The is not that much code in the sources, we could move most/all of it easily to the header files.

**Describe alternatives you've considered**
Alternatively we could leave the library as-is, but then it won't be integrated in the tensor-core correlator.

PR #220, opened 2023-09-22
Make cudawrappers header-only

csbnw:
**Description**

As described in #198, it would be better to have cudawrappers as a header-only library. This is realized in this PR.

Note that this is a breaking change and this PR should therefore only be merged after releasing version 0.5.0. Also, the changelog will be updated after rebasing this branch onto tag 0.5.0.

**Related issues**:

- https://github.com/nlesc-recruit/cudawrappers/issues/198


cmake/cudawrappers-config.cmake
```diff
@@ -0,0 +1,65 @@
...
+    if(${CUDAToolkit_VERSION_MAJOR} LESS ${CUDA_MIN_VERSION})
+      message(FATAL_ERROR "Insufficient CUDA version: " ${CUDAToolkit_VERSION}
+                          " < " ${CUDA_MIN_VERSION}
+      )
+    endif()
+  else()
+    message(
+      FATAL_ERROR
+        "CUDAToolkit not found... use find_package(CUDAToolkit REQUIRED)"
+    )
```
csbnw, line 20:
```suggestion
      FATAL_ERROR
        "CUDAToolkit not found, use find_package(CUDAToolkit REQUIRED)."
    )
```

john-romein:
I think that void DeviceMemory::zero(size_t size, Stream &stream) should be a member of Stream instead of a member of DeviceMemory.  Async copies are also a member of Stream, and an async zero is very similar to that.

csbnw:
> I think that `void DeviceMemory::zero(size_t size, Stream &stream)` should be a member of `Stream` instead of a member of `DeviceMemory`. Async copies are also a member of `Stream`, and an async zero is very similar to that.

This suggestion is outside the scope of this PR, but it should be fixed. To keep track of it, I created a new issue: https://github.com/nlesc-recruit/cudawrappers/issues/225

john-romein (approved):

PR #223, opened 2023-09-29
Fix kernel embedding

csbnw:
**Description**

`test_nvrtc` has a test case where the kernel source is embedded into the test source using a `include_cuda_code` helper function. This function doesn't work correctly when the source file is changed. It is now replaced with the `target_embed_source` helper function, which uses `ld` to link a static object with the code embedded. `test_nvrtc` is updated to demonstrate its use.

TODO:

- [x] Update `CHANGELOG.md`
- [ ] As soon as https://github.com/nlesc-recruit/cudawrappers/pull/220 is merged, export this function in the CMake config file so that it can also used by projects that use cudawrappers.

### DIFF
--- a/cmake/helper.cmake
+++ b/cmake/helper.cmake
@@ -1,7 +1,7 @@
 # Make it possible to embed a source file in a library, and link it to a target.
 # E.g. to link <kernel.cu> into target <example_program>, use
 # target_embed_source(example_program, kernel.cu). This will expose symbols
-# _binary_kernel_cu_start and _binary_ls_kernel_cu_end.
+# _binary_kernel_cu_start and _binary_kernel_cu_end.
 function(target_embed_source target input_file)
   # Strip the path and extension from input_file
   get_filename_component(NAME ${input_file} NAME_WLE)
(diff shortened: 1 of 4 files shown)


john-romein (approved):

cmake/helper.cmake
```diff
@@ -1,14 +1,23 @@
...
-  set(output_file "${CMAKE_CURRENT_BINARY_DIR}/${input_file}")
-  file(WRITE ${output_file} "${content}")
-  # Add save path to the include directories
-  get_filename_component(output_subdir ${input_file} DIRECTORY)
-  target_include_directories(
-    ${target} PRIVATE "${CMAKE_CURRENT_BINARY_DIR}/${output_subdir}"
+# Make it possible to embed a source file in a library, and link it to a target.
+# E.g. to link <kernel.cu> into target <example_program>, use
+# target_embed_source(example_program, kernel.cu). This will expose symbols
+# _binary_kernel_cu_start and _binary_ls_kernel_cu_end.
```
csbnw, line 4: Small typo:
```suggestion
# _binary_kernel_cu_start and _binary_kernel_cu_end.
```

PR #231, opened 2023-10-02
Fix variable not quoted can cause errors in other CMakefiles.txt

matmanc:
Because of an earlier change now in this CMakefile it is checked if the variable is defined. However, the variable is not properly escaped and it could cause issues.


csbnw (approved):
Good catch!

PR #236, opened 2023-10-09
Aartfaac fixes

john-romein:
**Description**

Adds cu::Context::getDevice() and fixes the cu::Module(CUmodule) constructor, necessary for use by the AARTFAAC correlator

**Related issues**:

- ...

### DIFF
--- a/include/cudawrappers/cu.hpp
+++ b/include/cudawrappers/cu.hpp
@@ -341,7 +341,7 @@ class Module : public Wrapper<CUmodule> {
     });
   }
 
-  explicit Module(Module &module) : Wrapper(module) {}
+  explicit Module(CUmodule &module) : Wrapper(module) {}
 
   CUdeviceptr getGlobal(const char *name) const {
     CUdeviceptr deviceptr{};
--- a/include/cudawrappers/cu.hpp
+++ b/include/cudawrappers/cu.hpp
@@ -90,6 +90,9 @@ class Device : public Wrapper<CUdevice> {
 
   explicit Device(int ordinal) { checkCudaCall(cuDeviceGet(&_obj, ordinal)); }
 
+  struct CUdeviceArg {}; // int and CUdevice are the same type, but we need two constructors
+  Device(CUdeviceArg, CUdevice device) : Wrapper(device) {}
+
   int getAttribute(CUdevice_attribute attribute) const {
     int value{};
     checkCudaCall(cuDeviceGetAttribute(&value, attribute, _obj));
@@ -192,6 +195,12 @@ class Context : public Wrapper<CUcontext> {
 
   void setSharedMemConfig(CUsharedconfig config) {
     checkCudaCall(cuCtxSetSharedMemConfig(config));
+  }
+
+  static Device getDevice() {
+    CUdevice device;
+    checkCudaCall(cuCtxGetDevice(&device));
+    return Device(Device::CUdeviceArg(), device);
   }
 
   static size_t getLimit(CUlimit limit) {
--- a/include/cudawrappers/cu.hpp
+++ b/include/cudawrappers/cu.hpp
@@ -90,7 +90,8 @@ class Device : public Wrapper<CUdevice> {
 
   explicit Device(int ordinal) { checkCudaCall(cuDeviceGet(&_obj, ordinal)); }
 
-  struct CUdeviceArg {}; // int and CUdevice are the same type, but we need two constructors
+  struct CUdeviceArg {
+  };  // int and CUdevice are the same type, but we need two constructors
   Device(CUdeviceArg, CUdevice device) : Wrapper(device) {}
 
   int getAttribute(CUdevice_attribute attribute) const {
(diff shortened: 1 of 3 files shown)


include/cudawrappers/cu.hpp
```diff
@@ -332,7 +342,7 @@ class Module : public Wrapper<CUmodule> {
     });
   }
 
-  explicit Module(Module &module) : Wrapper(module) {}
+  explicit Module(CUmodule &module) : Wrapper(module) {}
```
csbnw, line 345: This reverts the change made in https://github.com/nlesc-recruit/cudawrappers/pull/232, why is this needed?

include/cudawrappers/cu.hpp
```diff
@@ -194,6 +198,12 @@ class Context : public Wrapper<CUcontext> {
     checkCudaCall(cuCtxSetSharedMemConfig(config));
   }
 
+  static Device getDevice() {
```
csbnw, line 201: This new function should be mentioned in the `CHANGELOG.md`.

include/cudawrappers/cu.hpp
```diff
@@ -90,6 +90,10 @@ class Device : public Wrapper<CUdevice> {
 
   explicit Device(int ordinal) { checkCudaCall(cuDeviceGet(&_obj, ordinal)); }
 
+  struct CUdeviceArg {
+  };  // int and CUdevice are the same type, but we need two constructors
+  Device(CUdeviceArg, CUdevice device) : Wrapper(device) {}
+
```
csbnw, line 96: Who not simply?
```suggestion
  Device(int ordinal, CUdevice device) : Wrapper(device) {}

```

include/cudawrappers/cu.hpp
john-romein: Because these are totally different functions.  We do not need an explicit copy constructor (and definitely not one that has a non-const argument), but we do need a constructor that creates a Module from an existing CUmodule object.

include/cudawrappers/cu.hpp
john-romein: I do not want to document such minor changes twice.  We also have git log.

csbnw: This was not an optional request. I have now committed the changes to the `CHANGELOG.md` myself.

include/cudawrappers/cu.hpp
john-romein: These are also two different constructors.  One that takes a device ordinal as input, another one that takes a CUdevice as argument. They are subtly different: one is remapped through cuDeviceGet(); the other one not. Also, if Device would have had a destruction manager,  it would only be destroyed in the former case, as the ownership of the device would have remained by the user in the latter case.  Unfortunately both the device ordinal and CUdevice have "int" as the true type, so we need a method to distinguish both.

csbnw: So in case of `cuDeviceGet`, the wrapper is not responsible for destructing the `CUdevice`, while the new function takes ownership of the `CUdevice` and thus also needs to destruct it (hence the `Wrapper(device)`?

csbnw:
The new functions don't have unit test coverage, could you add (small) test cases to https://github.com/nlesc-recruit/cudawrappers/blob/main/tests/test_cu.cpp?

include/cudawrappers/cu.hpp
john-romein: It is the other way around.  If you would only use the CUDA wrappers, you need the constructor with the device ordinal as argument, which would (under the hood) be converted to a CUdevice through cuDeviceGet, and, as the Device resource is managed by the wrappers, would have had a manager if it would have had a destructor. (but it has no destructor so there is no need to manage it).  But if you already obtained a CUdevice object that you got someplace else and manage yourself, you can still wrap it in a Device object, but it is not managed by the wrappers (like any wrapped type), and as it is already a CUdevice object, it should not go trough cuDeviceGet (again).

csbnw:
> The new functions don't have unit test coverage, could you add (small) test cases to https://github.com/nlesc-recruit/cudawrappers/blob/main/tests/test_cu.cpp?

```
TEST_CASE("Test context::getDevice", "[device]") {
  cu::init();
  cu::Device device(cu::Context::getCurrent().getDevice());

  SECTION("Test Device.getName") {
    const std::string name = device.getName();
    std::cout << "Device name: " << name << std::endl;
    CHECK(name.size() > 0);
  }
}
```

This new simple test case fails:
```
FAILED:
due to unexpected exception with message:
  invalid device context
```

csbnw:
After an offline discussion with @john-romein I realize that `cu::Context::getCurrent().getDevice()` requires a pre-existing `cu::Device` and `cu::Context`. This is not clear when looking at the interface alone. A new test is added to verify this behaviour.

Maybe we should not add `Context::getDevice()` after all and wherever the device is needed simply pass a `cu::Device&` ?wherever

john-romein:
Well, the CUDA driver API exposes a function cuGetDevice, so it should be wrappable. Apparently, there are some restrictions on its use (which I do not fully understand), but any restriction on a driver API call also propagates to its wrapped equivalent, of course.
I will think about providing a Device argument explicitly when compiling through NVRTC; I'm not sure how it affects the applications. But that is a different issue.

csbnw (approved):

PR #239, opened 2023-10-25
Add helper function for embedding nvidiaMathDx

matmanc:
**Description**

In this pull request we adapt the cudawrappers to integrate easily the Nvidia MathDX library inside code.
In particular the decision was to embed the header only mathdx inside the executable and have the executable
make them available to the runtime compilation when needed.

### DIFF
--- a/cmake/cudawrappers-helper.cmake
+++ b/cmake/cudawrappers-helper.cmake
@@ -6,12 +6,18 @@ function(target_embed_source target input_file)
   include(CMakeDetermineSystem)
   # Strip the path and extension from input_file
   get_filename_component(NAME ${input_file} NAME_WLE)
+
+  set(WORK_DIR ${ARGN})
+
+  if(NOT DEFINED WORK_DIR)
+    set(WORK_DIR ${CMAKE_CURRENT_SOURCE_DIR})
+  endif()
   # Link the input_file into an object file
   add_custom_command(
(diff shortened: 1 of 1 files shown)


john-romein (commented):
Do we need MathDX support in the cudawrappers?  The cudawrappers are used on the host, and MathDX on the GPU.

matmanc:
This is a bit a question of convenience I think. These routines are only helping the user to embed the library so that is usable we are not writing an interface to them.

But I can also just remove the mathdx part.

cmake/cudawrappers-helper.cmake
```diff
@@ -6,12 +6,18 @@ function(target_embed_source target input_file)
   include(CMakeDetermineSystem)
   # Strip the path and extension from input_file
   get_filename_component(NAME ${input_file} NAME_WLE)
+
+  set(work_dir ${ARGN})
```
csbnw, line 10: Why not have a proper argument for the working directory?

csbnw:
After a discussion with @john-romein, we decided to close this PR. The newly added functionality doesn't fit the scope of this repository. By discarding this PR, the code is not lost, though, as it is used [here](https://git.astron.nl/RD/gpu-filter/-/blob/6e0fec3b32cd733efa805e2196c3fae278559f74/libfilter/CMakeLists.txt).

PR #240, opened 2023-11-07
Update cudawrappers to support COBALT

csbnw:
**Description**

A number of features were found missing when integrating cudawrappers into the COBALT code:
- Module constructor with CUjit_option map argument
- DeviceMemory::size
- HostMemory::size
- Function::launch with dim3 arguments
- Function::name
- Function::getAttribute is now const

**Related issues**:

- N.A.

### DIFF
--- a/include/cudawrappers/cu.hpp
+++ b/include/cudawrappers/cu.hpp
@@ -514,6 +514,13 @@ class Stream : public Wrapper<CUstream> {
                                  const_cast<void **>(&parameters[0]), nullptr));
   }
 
+  void launchKernel(Function &function, dim3 grid, dim3 block,
+                    unsigned sharedMemBytes,
+                    const std::vector<const void *> &parameters) {
+    launchKernel(function, grid.x, grid.y, grid.z, block.x, block.y, block.z,
+                 sharedMemBytes, parameters);
+  }
+
 #if CUDART_VERSION >= 9000
   void launchCooperativeKernel(Function &function, unsigned gridX,
                                unsigned gridY, unsigned gridZ, unsigned blockX,
--- a/include/cudawrappers/cu.hpp
+++ b/include/cudawrappers/cu.hpp
@@ -359,8 +359,8 @@ class Module : public Wrapper<CUmodule> {
       values.push_back(i.second);
     }
 
-    checkCudaCall(
-        cuModuleLoadDataEx(&_obj, image, options.size(), &keys[0], &values[0]));
+    checkCudaCall(cuModuleLoadDataEx(&_obj, image, options.size(), keys.data(),
+                                     values.data()));
 
     for (size_t i = 0; i < keys.size(); ++i) {
       options[keys[i]] = values[i];
--- a/include/cudawrappers/cu.hpp
+++ b/include/cudawrappers/cu.hpp
@@ -354,10 +354,9 @@ class Module : public Wrapper<CUmodule> {
     std::vector<CUjit_option> keys;
     std::vector<void *> values;
 
-    for (optionmap_t::const_iterator i = options.begin(); i != options.end();
-         ++i) {
-      keys.push_back(i->first);
-      values.push_back(i->second);
+    for (const std::pair<CUjit_option, void *> &i : options) {
+      keys.push_back(i.first);
+      values.push_back(i.second);
     }
 
     checkCudaCall(
--- a/include/cudawrappers/cu.hpp
+++ b/include/cudawrappers/cu.hpp
@@ -250,7 +250,7 @@ class Context : public Wrapper<CUcontext> {
 
 class HostMemory : public Wrapper<void *> {
  public:
-  explicit HostMemory(size_t size, unsigned int flags = 0) {
+  explicit HostMemory(size_t size, unsigned int flags = 0) : _size(size) {
     checkCudaCall(cuMemHostAlloc(&_obj, size, flags));
     manager = std::shared_ptr<void *>(new (void *)(_obj), [](void **ptr) {
       cuMemFreeHost(*ptr);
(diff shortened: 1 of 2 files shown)


include/cudawrappers/cu.hpp
```diff
@@ -342,6 +349,25 @@ class Module : public Wrapper<CUmodule> {
     });
   }
 
+  typedef std::map<CUjit_option, void *> optionmap_t;
+  explicit Module(const void *image, Module::optionmap_t &options) {
+    std::vector<CUjit_option> keys;
+    std::vector<void *> values;
+
+    for (optionmap_t::const_iterator i = options.begin(); i != options.end();
```
matmanc, line 357: you could use a for_each here https://en.cppreference.com/w/cpp/algorithm/for_each

include/cudawrappers/cu.hpp
```diff
@@ -342,6 +349,25 @@ class Module : public Wrapper<CUmodule> {
...
+         ++i) {
+      keys.push_back(i->first);
+      values.push_back(i->second);
+    }
+
+    checkCudaCall(
+        cuModuleLoadDataEx(&_obj, image, options.size(), &keys[0], &values[0]));
```
matmanc, line 364: why not here just .data() to get both pointer?

include/cudawrappers/cu.hpp
```diff
@@ -342,6 +349,25 @@ class Module : public Wrapper<CUmodule> {
...
+
+    for (size_t i = 0; i < keys.size(); ++i) {
+      options[keys[i]] = values[i];
```
matmanc, line 367: here you can have again the for each

include/cudawrappers/cu.hpp
```diff
@@ -342,6 +349,25 @@ class Module : public Wrapper<CUmodule> {
...
+    }
```
matmanc, line 368: Would it be an idea to add here a getFunction?

The module object has to be kept alive otherwise the function object becomes invalid

include/cudawrappers/cu.hpp
```diff
@@ -353,13 +379,13 @@ class Module : public Wrapper<CUmodule> {
 
 class Function : public Wrapper<CUfunction> {
  public:
-  Function(const Module &module, const char *name) {
+  Function(const Module &module, const char *name) : _name(name) {
     checkCudaCall(cuModuleGetFunction(&_obj, module, name));
   }
 
   explicit Function(CUfunction &function) : Wrapper(function) {}
```
matmanc, line 386: maybe remove the copy constructor or add a future deprecation warning.

The function cannot live without a module

include/cudawrappers/cu.hpp
```diff
@@ -509,6 +546,13 @@ class Stream : public Wrapper<CUstream> {
                                  const_cast<void **>(&parameters[0]), nullptr));
   }
 
+  void launchKernel(Function &function, dim3 grid, dim3 block,
```
matmanc, line 549: Isnt dim3 a defined in the cuda headers? This should be wrapped by cudawrapper so that when we make it cuda agnostic we dont change the underlining code. So maybe it is not a good idea to expose this interface. Maybe create a Dim3 wrapping type?

include/cudawrappers/cu.hpp
csbnw: Good suggestion, I applied it.

include/cudawrappers/cu.hpp
csbnw: How? `keys` and `values` are two distinct vectors.

include/cudawrappers/cu.hpp
csbnw: I don't see how this comment relates to the highlighted code.

include/cudawrappers/cu.hpp
csbnw: This is a topic for another discussion (outside the scope of this MR). I am merely storing the `name` as a member here and not altering the lifetime of the `module`.

include/cudawrappers/cu.hpp
csbnw: The COBALT code has a `struct Grid` and a `struct Block`. I don't really like it, but it is convenient. It is no big deal to use the more explicit cudawrappers interface and pass the `x`, `y`, and `z` values of the dimensions separately.

include/cudawrappers/cu.hpp
matmanc: you iterate on the map and capture keys and values in the lambda

include/cudawrappers/cu.hpp
csbnw: I updated the COBALT code to use the existing `launchKernel` (without `dim3`) and removed the newly added constructor.

include/cudawrappers/cu.hpp
matmanc: I mean to add it to the module class so I took an empty line in the code :D I see how that is confusing now

matmanc: We decide to keep this a separate topic in another future merge request

include/cudawrappers/cu.hpp
csbnw: I still don't understand how this could work. Can you provide the code?

include/cudawrappers/cu.hpp
csbnw: That is indeed nicer! (this code was copied and pasted without paying it much attention).

matmanc (approved):
Looks good to me! :)

include/cudawrappers/cu.hpp
csbnw: We discussed this on Slack and decided to stick to the current solution.

PR #244, opened 2024-01-03
Add asynchronous DeviceMemory::zero

csbnw:
**Description**

Added asynchronous `DeviceMemory::zero` with stream argument. This provides a wrapper for the `cuMemsetD8Async` API call.

**Related issues**:

- N.A.

### DIFF
--- a/include/cudawrappers/cu.hpp
+++ b/include/cudawrappers/cu.hpp
@@ -453,6 +453,10 @@ class DeviceMemory : public Wrapper<CUdeviceptr> {
     checkCudaCall(cuMemHostGetDevicePointer(&_obj, hostMemory, 0));
   }
 
+  void zero(size_t size, CUstream stream) {
+    checkCudaCall(cuMemsetD8Async(_obj, 0, size, stream));
+  }
+
   void zero(size_t size) { checkCudaCall(cuMemsetD8(_obj, 0, size)); }
 
   const void *parameter()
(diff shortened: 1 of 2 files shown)


include/cudawrappers/cu.hpp
```diff
@@ -453,6 +453,10 @@ class DeviceMemory : public Wrapper<CUdeviceptr> {
     checkCudaCall(cuMemHostGetDevicePointer(&_obj, hostMemory, 0));
   }
 
+  void zero(size_t size, CUstream stream) {
```
matmanc, line 456: I think this is already in the stream object. looks fine to me but you could just pass itself to the stream.

So something like stream.zero(_obj, size)

Also I am confused why zero needs a size since the size is saved in the memory object. Maybe an old interface issue.

the stream zero function has the interface (memory_pointer, size) here it is inverted you have (size, stream) maybe you could consider inverting the parameters?

csbnw: You are right, I missed the `Stream::zero` function completely. This function provides the same functionality as added in this MR, making this MR redundant.

There are cases where you don't want to zero the entire device memory. It would make sense to have the size as an optional argument, though, as indeed the `DeviceMemory::size()` could serve as default.

Regarding your last point, the newly added function is not needed after all, so the comment no longer applies.

I will now proceed with closing this MR.

PR #246, opened 2024-01-22
Support size = 0 for DeviceMemory constructor

csbnw:
**Description**

In some cases, a `cu::DeviceMemory` instance is needed of size zero. The CUDA driver API calls to allocate device memory don't support a zero size argument, so w need a special case in the constructor.

**Related issues**:

- N.A.


matmanc (approved):
Looks good

PR #247, opened 2024-01-22
Fix host memory

csbnw:
**Description**

The `cu::HostMemory` constructor passed the wrong pointer to `cuMemHostRegister`, which causes `cuMemHostUnregister` to fail. This happened silently, because `checkCudaCall` was missing. This guard was also missing in the other constructor, so added as well.

**Related issues**:

- N.A.


matmanc (commented):
looks good to me

matmanc (approved):
Looks good to me

PR #245, opened 2024-01-08
update links

laurasootes:
**Description**

This pull request fixes the deadlink issue in the documentation on readthedocs; the link to the contributing guidelines and readme dev. It uses the full main branch github links instead of the relative path. This is not the most elegant solution, but it does fix the links in the readthedocs which has not been working for quite some time.

**Related issues**:

- Fixes #133 

**Instructions to review the pull request**

- Check that CHANGELOG.md has been updated if necessary

PR #249, opened 2024-01-25
Add markdown link checker action

laurasootes:
**Description**

**Related issues**:

- #245 

I added a markdown link check action (for now only applied to this branch, should be changed to main branch before merging). However, it fails at the moment because codacy is not an open website; the link to it can only be opened with a password. Should we simply ignore this link in the check action, or try to fix it with the secrets?

Also, why is codacy not public?

### DIFF
--- a/.github/workflows/markdown-link-check.yml
+++ b/.github/workflows/markdown-link-check.yml
@@ -4,8 +4,7 @@ on:
   workflow_dispatch:
   push:
     branches:
-    #- main
-    - add_markdown_link_checker_action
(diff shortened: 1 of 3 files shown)


loostrum:
Nice addition! I think the codacy page should be public, no idea why it isn't.

loostrum:
@laurasootes what do we need before merging PR? Just the changing of the check to the main branch? I think we can ignore the codacy warning about readme.dev.md, that's just about line length

PR #251, opened 2024-02-09
Initialize size in Stream::memAllocAsync

csbnw:
**Description**

These changes fix the following bug: When calling `Stream::memAllocAsync`, a `DeviceMemory` is returned with it's `_size` member uninitialized. This is problematic when later calling `DeviceMemory::size`.

This is fixed by:
1. Deleting `DeviceMemory(CUdeviceptr)`
2. Adding `DeviceMemory(CUdeviceptr size)` and calling this constructor in `Stream::memAllocAsync`

A new test case was added to verify that the above scenario now works correctly

**Related issues**:

- N.A.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Fix compatibility with C++20 and C++23
 - Fix `cu::HostMemory` constructor for registered memory
 - Fix `cu::DeviceMemory` operator `T *()` for managed memory
+- Fix `Stream::memAllocAsync` returns `DeviceMemory` with initialized size
 
 ### Removed
 
--- a/include/cudawrappers/cu.hpp
+++ b/include/cudawrappers/cu.hpp
@@ -451,7 +451,10 @@ class DeviceMemory : public Wrapper<CUdeviceptr> {
                                            });
   }
 
-  explicit DeviceMemory(CUdeviceptr ptr) : Wrapper(ptr) {}
+  DeviceMemory(CUdeviceptr ptr) = delete;
+
+  explicit DeviceMemory(CUdeviceptr ptr, size_t size)
+      : Wrapper(ptr), _size(size) {}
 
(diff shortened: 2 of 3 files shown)


john-romein:
This change will break applications that use  DeviceMemory(CUdeviceptr ptr)  to cast an unwrapped pointer to a wrapped pointer.  We need this, because sometimes, pointer arithmetic (which can only be done on a CUdeviceptr) is necessary, e.g., to do a memcpyHtoD into the middle of a device memory block.
Maybe we should first rethink the semantics of a cu::DeviceMemory.  It used to used for both a true device memory block and a wrapped CUdeviceptr, but the addition of a "size" field only applies to former. Originally, the "size" field was added just for convenience, but now it drives us towards changing the semantics of the class, which needs a better thought. Possibly, we need two different classes (where the true memory block could inherit the properties of a wrapped pointer), but this needs a better thought as well.

csbnw:
> We need this, because sometimes, pointer arithmetic (which can only be done on a CUdeviceptr) is necessary, e.g., to do a memcpyHtoD into the middle of a device memory block.

I don't think that the pointer arithmetic should use `CUdeviceptr`, instead the `cu::DeviceMemory` should be cast to the actual data type (e.g. `float`).

> It used to used for both a true device memory block and a wrapped CUdeviceptr, but the addition of a "size" field only applies to former.
If the `cu::DeviceMemory` is initialized with a `CUdeviceptr` and the appropriate size, the `cu::DeviceMemory` class will work just fine.

Fixing this bug is pretty urgent in my opinion. A redesign of the `cu::DeviceMemory` (with different semantics) is beyond the scope of this PR. So @john-romein, how about we proceed with the proposed solution for now?

csbnw:
Alternatively, we could also decide to keep
```explicit DeviceMemory(CUdeviceptr ptr) : Wrapper(ptr) {}```
This won't break any existing code. The issue will still be fixed because of the newly added constructor.

john-romein:
Yes let's keep the original function too, otherwise it breaks the correlator applications.

john-romein (approved):

PR #255, opened 2024-03-04
The users should know that no libraries will be built.

### LINKED ISSUE #254: README says following build instructions will create a so, but a shared object is not built.
`libcudawrappers-*.so` is not built when executing
```
git clone https://github.com/nlesc-recruit/cudawrappers
cd cudawrappers
cmake -S . -B build
make -C build
```
After talking to @loostrum this is expected, since cudawrappers is now header only.

The README should be updated.

HannoSpreeuw:
The users should know that no libraries will be built, only headers will be produced.

The current README suggest that a `libcudawrappers-*.so` will appear in the build directory, which is not the case.

Fixes #254

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ cmake -S . -B build
 make -C build
 ```
 
-This command will create a `build` directory with a set of headers. Cudawrappers is header only, so no library objects are being built.
+This command will create a `build` directory. Cudawrappers is header only, so no library objects are being built.
 For more details on the building requirements and on testing, check the [developer documentation](README.dev.md).
 
 To install to `~/.local`, use
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ cmake -S . -B build
 make -C build
 ```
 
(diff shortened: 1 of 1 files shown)


csbnw:
Good point Hanno, the documentation was clearly outdated. However, there will be only some CMake configuration files in the build directory, no headers. How about removing the build part from the documentation altogether, and just mention that it is header-only?

README.md
```diff
@@ -43,7 +43,7 @@ cmake -S . -B build
 make -C build
 ```
 
-This command will create a `build` folder, compile the code and generate the libraries `libcudawrappers-*.so` in the build directory.
+This command will create a `build` directory with a set of headers. Cudawrappers is header only, so no library objects are being built.
```
csbnw, line 46:
```suggestion
This command will create a `build` directory. Cudawrappers is header only, so no library objects are being built.
```

csbnw (approved):
:+1:

PR #257, opened 2024-03-08
Release 0.7.0

csbnw:
**Description**

Update changelog to prepare for release 0.7.0

**Related issues**:

- N.A.


csbnw:
I noticed that despite the recently added `mdformat` linter, there were some inconsistencies with the escape characters in the version numbers. Also, `mdformat` modifies the list of diffs at the end, but doesn't include all versions. I have removed those diffs since they are not really needed anyway.

loostrum (approved):
Good catch on the square brackets in the version numbers

PR #266, opened 2024-04-24
Fix target_embed_source when input file is a relative path

csbnw:
**Description**

In some cases, `target_embed_source` fails when the input argument is a relative path that contains `..` . This is fixed by using `ABSOLUTE` instead of `WLE`. This also simplifies the call to `inline_local_includes`.

**Related issues**:

- N.A.


matmanc (approved):
Looks good!

PR #268, opened 2024-05-01
Extend api

csbnw:
**Description**

Extend the API so that cudawrappers can also be used in [PMT](https://git.astron.nl/RD/pmt):

- Added `cu::Device::getUUID()`
- Added `cudawrappers::nvml` target with only the functionality that PMT needs

**Related issues**:

- N.A.

### DIFF
--- a/include/cudawrappers/nvml.hpp
+++ b/include/cudawrappers/nvml.hpp
@@ -1,0 +1,50 @@
+#if !defined NVML_H
+#define NVML_H
+
+#include <nvml.h>
+
+#include <exception>
+
+#include <cudawrappers/cu.hpp>
+
+namespace nvml {
+class Error : public std::exception {
+ public:
+  explicit Error(nvmlReturn_t result) : _result(result) {}
+
+  const char* what() const noexcept { return nvmlErrorString(_result); }
+
+  operator nvmlReturn_t() const { return _result; }
+
+ private:
+  nvmlReturn_t _result;
+};
+
+inline void checkNvmlCall(nvmlReturn_t result) {
+  if (result != NVML_SUCCESS) throw Error(result);
+}
+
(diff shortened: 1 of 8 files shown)


include/cudawrappers/nvml.hpp
```diff
@@ -0,0 +1,51 @@
...
+};
+
+inline void checkNvmlCall(nvmlReturn_t result) {
+  if (result != NVML_SUCCESS) throw Error(result);
+}
+
+class Device {
+ public:
+  Device(int index) {
+    checkNvmlCall(nvmlInit());
```
matmanc, line 30: what happens if you init multiple times by creating 2 devices?

include/cudawrappers/nvml.hpp
```diff
@@ -0,0 +1,51 @@
...
+    checkNvmlCall(nvmlDeviceGetHandleByIndex(index, &device_));
+  }
+
+  Device(cu::Device& device) {
+    checkNvmlCall(nvmlInit());
```
matmanc, line 35: Same here

include/cudawrappers/nvml.hpp
csbnw: After giving it a quick test, calling `nvmlInit` multiple times doesn't seem to cause any problems.

csbnw:
After a discussion with @matmanc, we concluded that it is potentially dangerous if the `nvml::Device` initializes NVML and especially destroys it again (when other instances may still be active). A new `nvml::Context` class is added to guard against this. A `nvml::Context` must now be created first and next passed to the `nvml::Device` constructor. We also discussed adding `Context::getDevice`, but I think the current solution is cleaner.

matmanc (approved):
looks good

PR #270, opened 2024-05-02
Fix target embed source paths

loostrum:
**Description**

Fix target embed source linker input


csbnw (approved):
`input_file_inlined_relative` may be a bit ambiguous, relative to what?
But I think it's OK for now, we can always revisit this helper later and change all the variable names.

loostrum:
Yes I wasn't sure how to call that variable, a cleanup later may be good

PR #275, opened 2024-05-17
Fix string replacements in target_embed_source

csbnw:
**Description**

In some cases, the `target_embed_source` can fail. For instance, when using CMake's `FetchContent` feature. These changes fix it.

**Related issues**:

- N.A.


matmanc (approved):
Looks good ! :)

PR #279, opened 2024-06-07
Add inlined headers as dependencies in target_embed_source

loostrum:
This ensure that the embedded sources are rebuild when a header is changed.

### DIFF
--- a/cmake/cudawrappers-helper.cmake
+++ b/cmake/cudawrappers-helper.cmake
@@ -20,7 +20,10 @@ function(inline_local_includes input_file output_file)
   endforeach()
   string(APPEND processed_file_contents "${input_file_contents}\n")
   file(WRITE ${output_file} "${processed_file_contents}")
-  set(include_files ${include_files} PARENT_SCOPE)
+  set(include_files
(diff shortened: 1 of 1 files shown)


cmake/cudawrappers-helper.cmake
```diff
@@ -45,7 +51,7 @@ function(target_embed_source target input_file)
       ld ARGS -r -b binary -A ${CMAKE_SYSTEM_PROCESSOR} -o
       "${CMAKE_CURRENT_BINARY_DIR}/${NAME}.o" ${input_file_inlined_relative}
     WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
-    DEPENDS ${input_file} ${input_file_inlined}
+    DEPENDS ${input_file} ${input_file_inlined} ${include_files}
```
csbnw, line 54: I don't think that `${input_file_inlined}` is still needed here:
```suggestion
    DEPENDS ${input_file} ${include_files}
```

csbnw (approved):
One minor suggestion, other than that this looks like a nice new feature!

PR #290, opened 2024-07-18
Fix inline local includes edge cases

loostrum:
**Description**

- #include must be the start of a line
- the inlined included is placed in the same line as the original #include

**Related issues**:

This fixes cases such as
```
#ifdef __HIP__
//include "disabled_header.h"
#include "some_header.h"
#endif
```
Before this PR, the commented out include would be inlined, and both inlined headers would be placed at the top of the file.

### DIFF
--- a/cmake/cudawrappers-helper.cmake
+++ b/cmake/cudawrappers-helper.cmake
@@ -7,8 +7,10 @@ function(inline_local_includes input_file output_file)
   string(REGEX MATCHALL ${include_regex} includes ${input_file_contents})
   set(include_files "")
   foreach(include ${includes})
-    string(REGEX REPLACE ${include_regex} "\\3" include_name ${include}) # e.g. 'helper.h'
-    string(REGEX REPLACE ${include_regex} "\\2" include_line ${include}) # e.g.  '#include <helper.h>'
+    string(REGEX REPLACE ${include_regex} "\\3" include_name ${include}
+    )# e.g. 'helper.h'
+    string(REGEX REPLACE ${include_regex} "\\2" include_line ${include}
+    )# e.g.  '#include <helper.h>'
     file(GLOB_RECURSE INCLUDE_PATHS "${PROJECT_SOURCE_DIR}/*/${include_name}")
     if(NOT INCLUDE_PATHS STREQUAL "")
       list(SORT INCLUDE_PATHS ORDER DESCENDING)
--- a/cmake/cudawrappers-helper.cmake
+++ b/cmake/cudawrappers-helper.cmake
@@ -7,7 +7,7 @@ function(inline_local_includes input_file output_file)
   string(REGEX MATCHALL ${include_regex} includes ${input_file_contents})
   set(include_files "")
   foreach(include ${includes})
-    string(REGEX REPLACE ${include_regex} "\\3" include_name ${include})
+    string(REGEX REPLACE ${include_regex} "\\3" include_name ${include}) # e.g. 'helper.h'
     string(REGEX REPLACE ${include_regex} "\\2" include_line ${include})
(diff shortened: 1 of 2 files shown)


cmake/cudawrappers-helper.cmake
```diff
@@ -3,25 +3,24 @@
 # '#include "helper.h"` will lead to `helper.h` being inlined.
 function(inline_local_includes input_file output_file)
   file(READ ${input_file} input_file_contents)
-  set(include_regex "#include[ \t]*\"([^\"]+)\"")
+  set(include_regex "(^|\r?\n)(#include[ \t]*\"([^\"]+)\")")
```
loostrum, line 6: Note: the `?:` regex is not available, so the first match group will be the start of string / newline group. Hence in the rest of the function, group 2 is the entire `#include line` and group 3 is the header name

csbnw: I know that the existing function has no documentation inside the function body, but it may help to add it. I will make suggestions below.

cmake/cudawrappers-helper.cmake
```diff
@@ -3,25 +3,24 @@
...
   string(REGEX MATCHALL ${include_regex} includes ${input_file_contents})
-  string(REGEX REPLACE ${include_regex} "" input_file_contents
-                       "${input_file_contents}"
-  )
   set(include_files "")
   foreach(include ${includes})
-    string(REGEX REPLACE ${include_regex} "\\1" include ${include})
-    file(GLOB_RECURSE INCLUDE_PATHS "${PROJECT_SOURCE_DIR}/*/${include}")
+    string(REGEX REPLACE ${include_regex} "\\3" include_name ${include})
```
csbnw, line 10:
```suggestion
    string(REGEX REPLACE ${include_regex} "\\3" include_name ${include}) # e.g. 'helper.h'
```

cmake/cudawrappers-helper.cmake
```diff
@@ -3,25 +3,24 @@
...
+    string(REGEX REPLACE ${include_regex} "\\2" include_line ${include})
```
csbnw, line 11:
```suggestion
    string(REGEX REPLACE ${include_regex} "\\2" include_line ${include}) # e.g.  '#include <helper.h>'
```

csbnw (approved):
Nice fix! I added a small suggestion to document the new code.

cmake/cudawrappers-helper.cmake
```diff
@@ -7,8 +7,10 @@ function(inline_local_includes input_file output_file)
   string(REGEX MATCHALL ${include_regex} includes ${input_file_contents})
   set(include_files "")
   foreach(include ${includes})
-    string(REGEX REPLACE ${include_regex} "\\3" include_name ${include}) # e.g. 'helper.h'
-    string(REGEX REPLACE ${include_regex} "\\2" include_line ${include}) # e.g.  '#include <helper.h>'
+    string(REGEX REPLACE ${include_regex} "\\3" include_name ${include}
+    )# e.g. 'helper.h'
+    string(REGEX REPLACE ${include_regex} "\\2" include_line ${include}
+    )# e.g.  '#include <helper.h>'
```
csbnw, line 13: I don't like the auto-formatting in this case. How about changing it to:
```suggestion
    # Get the name of the file to include, e.g. 'helper.h'
    string(REGEX REPLACE ${include_regex} "\\3" include_name ${include})
    # Get the complete line of the include, e.g.  '#include <helper.h>'
    string(REGEX REPLACE ${include_regex} "\\2" include_line ${include})
```

csbnw:
Could you update the `CHANGELOG.md`, e.g. add something like "Made inlining of includes more robust" to the changes.

Issue #291, opened 2024-07-19
Add support for sliced DeviceMemory view


csbnw:
**Is your feature request related to a problem? Please describe.**
I would like to make a view of a `cu::DeviceMemory` object at some offset.

**Describe the solution you'd like**
Add a `cu::DeviceMemory` constructor that accepts both another `cu::DeviceMemory& other`, `size_t offset`, and `size_t size`.

**Describe alternatives you've considered**
Do this manually, using `static_cast<CUdeviceptr>` etc.

**Additional context**
N.A.

loostrum:
Added in #293

PR #293, opened 2024-07-19
Add option to create slice of device memory

loostrum:
**Description**

Added a `DeviceMemory` constructor that creates a slice from another `DeviceMemory` object, without taking ownership.
Tests for both correct and out-of-bounds usage were added.

**Related issues**:

#291


csbnw (approved):
Looks good, I have no remarks or suggestions: approved :+1:

Issue #296, opened 2024-07-23
Include cuda_runtime in cu.hpp


loostrum:
After merging the HIP branch, the main branch now no longer includes `cuda_runtime.h` in `cu.hpp`, see [here](https://github.com/nlesc-recruit/cudawrappers/blob/5098d73a8538d888351e3f08d0f7e11db57b77d6/include/cudawrappers/cu.hpp#L16-L19)
For HIP, the hip runtime _is_ included. Should we include `cuda_runtime.h` (or perhaps `cuda_runtime_api.h`) in `cu.hpp`?
Existing code in e.g. `ccglib` is now broken with CUDA because `dim3` no longer exists when only including cudawrappers.

csbnw:
Since `cuda_runtime_api.h` was there before, I indeed agree that it should be restored.

loostrum:
Fixed in #297

PR #297, opened 2024-07-23
include cuda_runtime in cu.hpp

loostrum:
**Description**

Include `cuda_runtime.h` in `cu.hpp`.

**Related issues**:
See #296 for a description of the problem this PR solves.

### DIFF
--- a/include/cudawrappers/cu.hpp
+++ b/include/cudawrappers/cu.hpp
@@ -15,6 +15,7 @@
 
 #if !defined(__HIP__)
 #include <cuda.h>
+#include <cuda_runtime.h>
 #else
 #include <hip/hip_runtime.h>
 


include/cudawrappers/cu.hpp
```diff
@@ -14,6 +14,8 @@
 #include <vector>
 
 #if !defined(__HIP__)
+#include <cuda_runtime.h>
+
 #include <cuda.h>
```
csbnw, line 19: Reorder to put them alphabetically, also, the newline is not needed:
```suggestion
#include <cuda.h>
#include <cuda_runtime.h>
```

csbnw (changes requested):

include/cudawrappers/cu.hpp
loostrum: I had them in the wrong order indeed, not sure why pre-commit decided to add a newline instead of switching them around...

loostrum:
Pre-commit reverted your change @csbnw, perhaps it doesn't like the underscore...

csbnw (approved):
I am not sure what pre-commit is up to. I don't like its auto-formatting in this case, but I'll approve this PR anyway.

loostrum:
Perhaps we can fiddle with some of pre-commit's settings to avoid weird issues like this. Weirdly it did pass locally (on my AMD machine, on DAS6 pre-commit doesn't work at all anymore)

PR #301, opened 2024-08-20
Update target_embed_source to re-run when the sources change

csbnw:
**Description**

The `target_embed_source` helper function didn't properly implement dependency management. This is now changed so that whenever either the kernel source file itself, or any of the local header files that in includes change, a new inlined header file is written and compiled into an object file.

The solution was to use a second `add_custom_command` call used to do the inlining. Using the `DEPENDS` option, dependencies can be specified so that the command runs again when any of them change. This has two consequences:

1. The `inline_local_includes` function had to be moved to a separate file (`cudawrappers-inline-local-includes.cmake`), since the `COMMAND` option for `add_custom_command` can not run a CMake function directly :facepalm: 
2. The part of `inline_local_includes` that retrieves the list of absolute file names for the local include directories had to be replicated (it is called `get_local_includes`) so that we can pass that list to the `DEPENDS` option. Passing the list of paths to `inline_local_includes` to avoid replicated code was considered, but this will only make the code more complicated.

**Related issues**:

- N.A.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,6 +20,7 @@ repos:
     hooks:
       - id: cmake-format
       - id: cmake-lint
+        args: ['--line-width=90']
   - repo: https://github.com/executablebooks/mdformat
     rev: 0.7.17
     hooks:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
   includes and respects the location of an include in the original source file
 - Upgrade C++ standard to C++14
 - Upgrade Catch2 to version v3.6.0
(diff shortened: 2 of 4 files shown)


matmanc (commented):
looks good. I see your are getting the root_dir to the command and then back as an argument. I guess that is the only way or have you considered other ways?

matmanc (approved):

csbnw:
> looks good. I see your are getting the root_dir to the command and then back as an argument. I guess that is the only way or have you considered other ways?

This seems to be the simplest way to get the desired behaviour in CMake. You have to call a `.cmake` file, so there should be a function call somewhere in that file, hence `inline_local_includes(${input_file} ${output_file} ${root_dir})`. It works in the same way for the `input_file` and `output_file` arguments.

PR #302, opened 2024-09-24
Add nvml::Device::getPower

csbnw:
**Description**

Support for `nvmlFieldValue_t` was added in CUDA 12 and can (amongst other things) be used to query power usage for a CUDA device. In earlier CUDA versions (e.g. CUDA 11.8), it was common to use `nvmlDeviceGetPowerUsage`, but it is missing from the API. This function has therefore been added to the `nvml::Device` class as `getPower` so that one can query power usage even when using older versions of CUDA.

**Related issues**:

- N.A.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
   `cu::DeviceMemory` object
 - Added `cu::DeviceMemory::memset()`
 - Added `cu::Stream::memsetAsync()`
+- Added `nvml::Device::getPower()`
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


loostrum (commented):
Nice addition, is it worth adding a test for this new functionality?

csbnw:
> Nice addition, is it worth adding a test for this new functionality?
That is a good idea, a test is added.

Note that the unit of power is not mentioned in cudawrappers. Like for the other wrapped API calls, users should refer to the [NVIDIA documentation](https://docs.nvidia.com/deploy/nvml-api/group__nvmlDeviceQueries.html#group__nvmlDeviceQueries_1g7ef7dff0ff14238d08a19ad7fb23fc87) for details.

loostrum (approved):

PR #305, opened 2024-10-31
Added support for 2D operations and C2R/R2C FFT

wvbbreu:
**Description**

This pull request includes support for various 2D operations, namely `memset2D`, `memcpyHtoD2DAsync`, `memcpyDtoH2DAsync,` and introduces `FFT1DRealToComplex` and `FFT1DComplexToReal` classes for FFT transformations. Finally, the `cu::Device::getOrdinal()` method is introduced to retrieve the current device ID.
Related tests are included in this pull request.

**Related issues**:

None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,12 @@ project adheres to [Semantic Versioning](http://semver.org/).
 - Added `cu::DeviceMemory::memset()`
 - Added `cu::Stream::memsetAsync()`
 - Added `nvml::Device::getPower()`
-- Added 2D memcpy and memset operations
-- Added `FFT1DRealToComplex` and `FFT1DComplexToReal`
+- Added `cu::Stream::memcpyHtoD2DAsync()`, `cu::Stream::memcpyDtoHD2Async()`,
+  and `cu::Stream::memcpyDtoD2DAsync()` for 2D asynchronous memory copies.
+- Added `cu::DeviceMemory::memset2D()` and `cu::Stream::memset2DAsync()` for 2D
+  memsets
+- Added `cufft::FFT1D_R2C` and `cufft::FFT1D_C2R` for 1D real-to-complex and
+  vice verse FFT
 
 ### Changed
 
@@ -26,6 +30,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 - Upgrade Catch2 to version v3.6.0
 - `target_embed_source` is now more robust: it properly tracks dependencies and
   runs again whenever any of them changes
+- Expanded tests to cover the new 2D memory operations and FFT support
 
 ## \[0.8.0\] - 2024-07-05
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 - Added `cu::DeviceMemory::memset()`
 - Added `cu::Stream::memsetAsync()`
 - Added `nvml::Device::getPower()`
+- Added 2D memcpy and memset operations
+- Added `FFT1DRealToComplex` and `FFT1DComplexToReal` 
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 - Added `cu::Stream::memsetAsync()`
 - Added `nvml::Device::getPower()`
 - Added 2D memcpy and memset operations
-- Added `FFT1DRealToComplex` and `FFT1DComplexToReal` 
+- Added `FFT1DRealToComplex` and `FFT1DComplexToReal`
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
   memsets
 - Added `cufft::FFT1D_R2C` and `cufft::FFT1D_C2R` for 1D real-to-complex and
   vice versa FFT
+- Added cu::Device::getOrdinal().
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 - Added `cu::Stream::memcpyHtoD2DAsync()`, `cu::Stream::memcpyDtoHD2Async()`,
   and `cu::Stream::memcpyDtoD2DAsync()`
 - Added `cu::DeviceMemory::memset2D()` and `cu::Stream::memset2DAsync()`
-- Added `cufft::FFT1D_R2C` and `cufft::FFT1D_C2R`
+- Added `cufft::FFT1DR2C` and `cufft::FFT1DC2R`
 - Added `cu::Device::getOrdinal()`
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,12 +15,10 @@ project adheres to [Semantic Versioning](http://semver.org/).
 - Added `cu::Stream::memsetAsync()`
 - Added `nvml::Device::getPower()`
 - Added `cu::Stream::memcpyHtoD2DAsync()`, `cu::Stream::memcpyDtoHD2Async()`,
-  and `cu::Stream::memcpyDtoD2DAsync()` for 2D asynchronous memory copies.
-- Added `cu::DeviceMemory::memset2D()` and `cu::Stream::memset2DAsync()` for 2D
-  memsets
-- Added `cufft::FFT1D_R2C` and `cufft::FFT1D_C2R` for 1D real-to-complex and
-  vice versa FFT
-- Added cu::Device::getOrdinal().
+  and `cu::Stream::memcpyDtoD2DAsync()`
+- Added `cu::DeviceMemory::memset2D()` and `cu::Stream::memset2DAsync()`
+- Added `cufft::FFT1D_R2C` and `cufft::FFT1D_C2R`
+- Added `cu::Device::getOrdinal()`
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 - Added `cu::DeviceMemory::memset2D()` and `cu::Stream::memset2DAsync()` for 2D
   memsets
 - Added `cufft::FFT1D_R2C` and `cufft::FFT1D_C2R` for 1D real-to-complex and
-  vice verse FFT
+  vice versa FFT
 
 ### Changed
 
--- a/include/cudawrappers/cu.hpp
+++ b/include/cudawrappers/cu.hpp
@@ -23,7 +23,6 @@
 #include <cudawrappers/macros.hpp>
 #endif
 
-
 namespace cu {
 class Error : public std::exception {
  public:
@@ -190,7 +189,8 @@ class Device : public Wrapper<CUdevice> {
   }
 
   size_t getTotalConstMem() const {
-    return static_cast<size_t>(getAttribute(CU_DEVICE_ATTRIBUTE_TOTAL_CONSTANT_MEMORY));
+    return static_cast<size_t>(
+        getAttribute(CU_DEVICE_ATTRIBUTE_TOTAL_CONSTANT_MEMORY));
   }
 
   // Primary Context Management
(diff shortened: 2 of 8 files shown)


include/cudawrappers/nvtx.hpp
```diff
@@ -1,14 +1,20 @@
 #if !defined NVTX_H
 #define NVTX_H
 
+#if !defined(__HIP__)
 #include <nvToolsExt.h>
+#endif
```
csbnw, line 6: The `Marker` class doesn't do anything when compiling with HIP,  but one doesn't have to change their code if they used `nvtx::Marker`. I like it! In a follow up MR, the functionallity could be implemented using [rocxt](https://rocm.docs.amd.com/projects/roctracer/en/latest/reference/roctx-spec.html).

tests/test_cu.cpp
```diff
@@ -23,6 +23,20 @@ TEST_CASE("Test cu::Device", "[device]") {
     std::cout << "Device arch: " << arch << std::endl;
     CHECK(arch.size() > 0);
   }
+
+  SECTION("Test Device.totalMem") {
```
csbnw, line 27: For consistency:
```suggestion
  SECTION("Test device::totalMem", "[device]") {
```

tests/test_cu.cpp
```diff
@@ -23,6 +23,20 @@ TEST_CASE("Test cu::Device", "[device]") {
...
+    const size_t total_mem = device.totalMem();
+    std::cout << "Device total memory: " << (total_mem / (1024 * 1024))
+              << " bytes" << std::endl;
+    CHECK(total_mem > 0);
+  }
+
+  SECTION("Test Device.getTotalConstMem") {
```
csbnw, line 34: Same here, update so that the section names matches the API that is being tested.

tests/test_cufft.cpp
```diff
@@ -114,6 +114,34 @@ TEST_CASE("Test 1D FFT", "[FFT1D]") {
     scaleSignal(out_ptr, out_ptr, size, float(size));
     compare(out_ptr, in_ptr, size);
   }
+
+  SECTION("FP32 R2C C2R") {
```
csbnw, line 118: This section name is inconsistent compared to the other names in `test_cufft.cpp`.

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 - Added `cu::DeviceMemory::memset()`
 - Added `cu::Stream::memsetAsync()`
 - Added `nvml::Device::getPower()`
+- Added 2D memcpy and memset operations
+- Added `FFT1DRealToComplex` and `FFT1DComplexToReal`
```
csbnw, line 18: See the other 'Added' lines for the preferred format. Name the namespace and device function exactly as they appear in the header file.

include/cudawrappers/cufft.hpp
```diff
@@ -224,6 +225,85 @@ FFT2D<CUDA_C_16F>::FFT2D(int nx, int ny, int stride, int dist, int batch) {
...
+                     long long ouembed) = delete;
+};
+
+template <>
+FFT1DComplexToReal<CUDA_C_32F>::FFT1DComplexToReal(int nx, int batch,
+                                                   long long int inembed,
+                                                   long long int ouembed) {
+  checkCuFFTCall(cufftCreate(plan()));
+  const int rank = 1;
+  size_t ws = 0;
```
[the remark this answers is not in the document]
wvbbreu: `cufftXtMakePlanMany`requires a pointer to `ws` of type `int`.

include/cudawrappers/cufft.hpp
```diff
@@ -224,6 +225,85 @@ FFT2D<CUDA_C_16F>::FFT2D(int nx, int ny, int stride, int dist, int batch) {
...
+#if defined(__HIP__)
+  __host__
+#endif
+  FFT1DComplexToReal(int nx, int batch) = delete;
+#if defined(__HIP__)
+  __host__
+#endif
+  FFT1DComplexToReal(int nx, int batch, long long inembed,
+                     long long ouembed) = delete;
+};
```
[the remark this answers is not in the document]
wvbbreu: Good suggestion, I agree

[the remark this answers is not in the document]
csbnw: I agree that `FFT1D_C2R` is easier to read, but using underscores is against [Google Style](https://google.github.io/styleguide/cppguide.html#Type_Names) (which we try to adhere to).

CHANGELOG.md
```diff
@@ -14,6 +14,13 @@ project adheres to [Semantic Versioning](http://semver.org/).
 - Added `cu::DeviceMemory::memset()`
 - Added `cu::Stream::memsetAsync()`
 - Added `nvml::Device::getPower()`
+- Added `cu::Stream::memcpyHtoD2DAsync()`, `cu::Stream::memcpyDtoHD2Async()`,
+  and `cu::Stream::memcpyDtoD2DAsync()` for 2D asynchronous memory copies.
+- Added `cu::DeviceMemory::memset2D()` and `cu::Stream::memset2DAsync()` for 2D
+  memsets
+- Added `cufft::FFT1D_R2C` and `cufft::FFT1D_C2R` for 1D real-to-complex and
+  vice versa FFT
+- Added cu::Device::getOrdinal().
```
csbnw, line 23: No need to elaborate (this can optionally be done in the release notes), the markdown formatting on the last new line was missing:
```suggestion
- Added `cu::Stream::memcpyHtoD2DAsync()`, `cu::Stream::memcpyDtoHD2Async()`,
  and `cu::Stream::memcpyDtoD2DAsync()`
- Added `cu::DeviceMemory::memset2D()` and `cu::Stream::memset2DAsync()`
- Added `cufft::FFT1D_R2C` and `cufft::FFT1D_C2R`
- Added `cu::Device::getOrdinal()`
```

include/cudawrappers/cu.hpp
```diff
@@ -182,12 +182,19 @@ class Device : public Wrapper<CUdevice> {
...
   }
 
-  size_t totalMem() const {
+  size_t getTotalMem() const {
     size_t size{};
     checkCudaCall(cuDeviceTotalMem(&size, _obj));
     return size;
   }
 
+  size_t getTotalConstMem() const {
```
csbnw, line 191: This changes the API. Moreover, we try to stick to the names of the CUDA API's. Since it uses `cuDeviceTotalMem` and not `cuDeviceGetTotalMem`, `totalMem` is a good name for this function.

Regarding `getTotalConstMem`, there is already the `cu::Device::getAttribute` function. No need to add a function for individual attributes, so I suggest removing `getTotalConstMem`.

include/cudawrappers/cu.hpp
```diff
@@ -597,13 +599,40 @@ class DeviceMemory : public Wrapper<CUdeviceptr> {
...
+  }
+
   void zero(size_t size) { memset(static_cast<unsigned char>(0), size); }
 
   const void *parameter()
       const  // used to construct parameter list for launchKernel();
   {
     return &_obj;
   }
+  void *parameter_copy() { return reinterpret_cast<void *>(_obj); }
```
csbnw, line 635: This function doesn't seem to be used, remove it?
```suggestion
```

wvbbreu: This function was intentionally added to demonstrate its use through a small example. In some libraries, such as [dedisp](http://git.astron.nl/rd/dedisp), a significant amount of manual pointer arithmetic is (unfortunately) performed. Using the dereference operator (e.g., `cu::Device::operation*`) would be quite beneficial in such cases. However, it's not allowed, as memory in these situations isn’t allocated with `CU_POINTER_ATTRIBUTE_IS_MANAGED`, resulting in a runtime exception.

## Option 1 (Dereference Operator)
```cpp
cu::DeviceMemory idata(...); // idata points to a device address

unsigned int i = 1;
unsigned int nsamp_padded = 2345;
unsigned int nchan_fft_batch = 5678;
cu::DeviceMemory idata(static_cast<cufftReal*>(d_data_x_nu) + i * nsamp_padded * nchan_fft_batch);
// --> THROWS: "Cannot return memory of type CU_MEMORYTYPE_DEVICE as pointer."
```

## Option 2 (`cu::DeviceMemory` Constructor with Offset)
Alternatively, the cu::DeviceMemory(DeviceMemory&, size_t offset, size_t size) constructor allows us to wrap the arithmetic within the DeviceMemory object itself. Here, offset is given in bytes, so we need to multiply by the word size (e.g., sizeof(cufftReal)). Even so, I’d prefer to do pointer arithmetic with a cufftReal pointer rather than manually determining the byte offset.
```cpp
cu::DeviceMemory idata(d_data_x_nu, (i * nsamp_padded * nchan_fft_batch) * sizeof(cufftReal), d_data_x_nu.size());
```

## Option 3 (`cu::DeviceMemory::parameter`)
As an alternative, we could use cu::DeviceMemory::parameter(). However, it has its limitations. This function returns a const pointer to the _obj pointer stored in cu::DeviceMemory, not the actual underlying device pointer, so it can’t be used directly. Personally, I would rather avoid using this type of extensive casting in production.

```cpp
const void* const* param = reinterpret_cast<const void* const*>(d_data_x_nu.parameter());

cu::DeviceMemory idata(reinterpret_cast<CUdeviceptr>(
    static_cast<const cufftReal*>(*param) +
    i * nsamp_padded * nchan_fft_batch));
```

## Option 4 (`cu::DeviceMemory::parameter_copy`)
A final approach is to return the actual device pointer by introducing a `parameter_copy()` method.

```cpp
cu::DeviceMemory idata(static_cast<const cufftReal*>(d_data_x_nu.parameter_copy()) + i * nsamp_padded * nchan_fft_batch);
```

csbnw: Option 4 looks much cleaner than option 3, but I really think that the name `parameter_copy` should be changed. It shouldn't contain an underscore (see also [this](https://google.github.io/styleguide/cppguide.html#Function_Names) page) but more importantly, `copy` is somewhat misleading. I think `object` or `pointer` would be better names.

Another thing that comes to mind is that a feature like this may be useful for the other classes in `cu::`, for instance when combining cudawrappers code with 'plain' CUDA code. So this function could even be added to `cu::Wrapper`?

@loostrum, @john-romein, what is your view here?

john-romein: Please stay as close as possible to the CUDA driver API.  If you need to do arithmetic on a device memory pointer, simply cast the cu::DeviceMemory object to a CUdeviceptr (a cast operator that is inherited from cu::Wrapper<CUdeviceptr>) and do the arithmetic on CUdeviceptr (which is some int type that allows arithmetic to be performed on).

wvbbreu: I had not considered that approach, but it does work. The only downside is that for non-char sized types, it requires two casts. For example:

```cpp
float* x = reinterpret_cast<float*>(static_cast<CUdeviceptr>(some_dev_object));
```

@john-romein @csbnw Would we rather opt for this approach than introducing a new method?

wvbbreu: The problem with performing arithmetic on a `CUdeviceptr` is that, in HIP, the underlying type is `void*` (`hipDeviceptr_t`), which does not support pointer arithmetic (see compiler warning). In contrast, CUDA allows this because the underlying type is `unsigned long long`.

```
error: arithmetic on a pointer to void
  549 |             static_cast<CUdeviceptr>(some_dev_object) + (3 * sizeof(float))
```

This issue will likely also exist in other applications that use pointer arithmetic and are not yet compiled with HIP.

wvbbreu:
I came up with a solution that relaxes the requirements of `cu::DeviceMemory::operator*` a bit. Instead of only allowing de-referencing on managed memory, access to non-managed/device memory is also allowed (see example). This change as no impact on the existing cudawrappers API. Access to unallocated memory will still be handled through a sanity check (`checkPointerAccess`).

## Before
```cpp
cu::DeviceMemory(1024) dev_mem;
cu::DeviceMemory(reinterpret_cast<CUdeviceptr>(static_cast<float*>(static_cast<CUdeviceptr>(dev_mem) + 2));
```

## After
```cpp
cu::DeviceMemory(1024) dev_mem;
cu::DeviceMemory(reinterpret_cast<CUdeviceptr>(static_cast<float*>(dev_mem) + 2));
```

include/cudawrappers/cu.hpp
```diff
@@ -94,6 +94,22 @@ class Wrapper {
...
   explicit Wrapper(T &obj) : _obj(obj) {}
 
+  template <CUmemorytype... AllowedMemoryTypes>
+  inline void checkPointerAccess(const CUdeviceptr &pointer) const {
+    CUmemorytype memoryType;
+    checkCudaCall(cuPointerGetAttribute(
+        &memoryType, CU_POINTER_ATTRIBUTE_MEMORY_TYPE, pointer));
+
+    // Check if the memoryType is one of the allowed memory types
+    bool isAllowed = false;
```
csbnw, line 104: Not used any more, remove it:
```suggestion
```

include/cudawrappers/cu.hpp
```diff
@@ -94,6 +94,22 @@ class Wrapper {
...
+    for (auto allowedType : {AllowedMemoryTypes...}) {
+      if (memoryType == allowedType) return;
```
csbnw, line 106: Google Style rule suggests adding curly braces to improve readability and avoid errors related to ambiguous indentation or adding lines later:
```suggestion
      if (memoryType == allowedType) {
            return;
      }
```

csbnw (commented):
Two minor suggestions, other than that I think it's ready to be merged.

csbnw (approved):

PR #312, opened 2025-01-24
Add graph support

matmanc:
CUDA Graph API seems a promising addition to the already existing cuda streams. Especially to mix host and device function in a complex order. Moreover, also HIP supports cuda graphs so it feels natural to extend cudawrappers to supports graph.

In this pull request the basic functionality of cuda graph api are introduced.

Clone and verify on a machine with an NVIDIA GPU
```
git clone https://github.com/nlesc-recruit/cudawrappers .
git checkout <this-branch>
cmake -S . -B build -CD -DCUDAWRAPPERS_BUILD_TESTING=True
make -C build
make -C build format # Nothing should change
make -C build lint # linting is broken until we fix the issues (see #92)
build/tests/test_graph
```

Clone and verify on a machine with HIP and AMD GPU
```
git clone https://github.com/nlesc-recruit/cudawrappers .
git checkout <this-branch>
cmake -S . -B build -CD -DCUDAWRAPPERS_BUILD_TESTING=True -DCUDA_WRAPPERS_BACKEND_HIP=True
make -C build
make -C build format # Nothing should change
build/tests/test_graph
```


include/cudawrappers/cu.hpp
```diff
@@ -668,6 +668,248 @@ class DeviceMemory : public Wrapper<CUdeviceptr> {
...
+  GraphKernelNodeParams(const Function &function, unsigned gridDimX,
+                        unsigned gridDimY, unsigned gridDimZ,
+                        unsigned blockDimX, unsigned blockDimY,
+                        unsigned blockDimZ, unsigned sharedMemBytes,
+                        const std::vector<const void *> &params) {
+    _obj.func = function;
+#if defined(__HIP__)
+    _obj.blockDim = {blockDimX, blockDimY, blockDimZ};
+    _obj.gridDim = {gridDimX, gridDimY, gridDimZ};
+
```
csbnw, line 692:
```suggestion
```

include/cudawrappers/cu.hpp
```diff
@@ -668,6 +668,248 @@ class DeviceMemory : public Wrapper<CUdeviceptr> {
...
+                   const std::vector<CUgraphNode> &dependencies,
+                   GraphHostNodeParams &params) {
+    checkCudaCall(cuGraphAddHostNode(
+        node.getNode(), _obj, dependencies.data(), dependencies.size(),
+        reinterpret_cast<CUDA_HOST_NODE_PARAMS *>(&params)));
+  }
+
+  void addDevMemFreeNode(GraphNode &node,
+                         const std::vector<CUgraphNode> &dependencies,
+                         const CUdeviceptr &devPtr) {
```
csbnw, line 836: Shouldn't this be a `cu::DeviceMemory&`?

include/cudawrappers/cu.hpp
```diff
@@ -668,6 +668,248 @@ class DeviceMemory : public Wrapper<CUdeviceptr> {
...
+  }
+
+  void addHostToDeviceMemCopyNode(GraphNode &node,
+                                  const std::vector<CUgraphNode> &dependencies,
+                                  GraphMemCopyToDeviceNodeParams &params,
+                                  Context &ctx) {
+#if defined(__HIP__)
+    /*checkCudaCall(hipGraphAddMemcpyNode(
+        node.getNode(), _obj, dependencies.data(), dependencies.size(),
+        reinterpret_cast<hipMemcpy3DParms *>(&params)));*/
```
csbnw, line 857:
```suggestion
```

include/cudawrappers/cu.hpp
```diff
@@ -668,6 +668,248 @@ class DeviceMemory : public Wrapper<CUdeviceptr> {
...
+#endif
+    _obj.sharedMemBytes = sharedMemBytes;
+    _obj.kernelParams = const_cast<void **>(params.data());
+    _obj.extra = nullptr;
+  }
+};
+
+class GraphHostNodeParams : public Wrapper<CUDA_HOST_NODE_PARAMS> {
+ public:
+  GraphHostNodeParams(void (*fn)(void *), void *data) {
```
wvbbreu, line 708: You could define `fn` as a `std::function` instead of a C-style function pointer. What do you think?

include/cudawrappers/cu.hpp
```diff
@@ -668,6 +668,248 @@ class DeviceMemory : public Wrapper<CUdeviceptr> {
...
+    checkCudaCall(cuGraphAddMemcpyNode(
+        node.getNode(), _obj, dependencies.data(), dependencies.size(),
+        reinterpret_cast<CUDA_MEMCPY3D *>(&params), ctx));
+#endif
+  }
+
+  void exportDotFile(
+      std::string path,
+      CUgraphDebugDot_flags flags =
+          CUgraphDebugDot_flags::CU_GRAPH_DEBUG_DOT_FLAGS_VERBOSE) {
```
csbnw, line 890: How about making this a function argument with default value?

tests/test_graph.cpp
```diff
@@ -0,0 +1,192 @@
...
+    cu::GraphExec graph_exec(graph);
+    cu::Stream stream;
+    stream.lunchGraph(graph_exec);
+    stream.synchronize();
+
+    CHECK(data == 44);
+  }
+  SECTION("Test cu::Graph : memory management") {
+    std::vector<float> data_in(1000);   //{3, 3, 3};
+    std::vector<float> data_out(1000);  //{0, 0, 0};
```
csbnw, line 57:
```suggestion
    std::vector<float> data_in(1000);
    std::vector<float> data_out(1000);
```

tests/test_graph.cpp
```diff
@@ -0,0 +1,192 @@
...
+  nvrtc::Program program(kernel, "kernel.cu");
+  program.compile({});
+  cu::Module module(static_cast<const void*>(program.getPTX().data()));
+
+  cu::Stream stream;
+
+  SECTION("Test cu::hostNode single value") {
+    auto fn = [](void* data) {
+      int* ptr = static_cast<int*>(data);
+      std::cout << "Host node content: " << *ptr << std::endl;
```
csbnw, line 37: Put this inside a `#ifdef DEBUG` guard, or remove it all? The tests will mostly be ran automated, no-one will look at those print statements.

tests/test_graph.cpp
```diff
@@ -0,0 +1,192 @@
...
+#include <iostream>
+#include <string>
+
+#include <cudawrappers/cu.hpp>
+#include <cudawrappers/nvrtc.hpp>
+
+TEST_CASE("Test cu::Graph", "[graph]") {
+  const std::string kernel = R"(
+    
+
```
csbnw, line 14:
```suggestion
```

tests/test_graph.cpp
```diff
@@ -0,0 +1,192 @@
...
+    cu::Graph graph;
+    cu::GraphNode dev_alloc, host_set, copy_to_dev, execute_kernel, device_free,
+        copy_to_host, host_set2;
+
+    auto set_value = [](void* data) {
+      float* ptr = static_cast<float*>(data);
+      for (int i = 0; i < 3; i++) {
+        ptr[i] = 42.0f;
+      }
+    };
```
csbnw, line 148: Add a constant `n = 3` to avoid duplicating the length?

tests/test_graph.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+    cu::GraphMemCopyToHostNodeParams copy_to_host_params{
+        data_out.data(),
+        dev_alloc_params.getDevPtr(),
+        sizeof(data_in[0]),
+        data_in.size(),
+        1,
+        1};
+
+    size_t array_size = 3;
```
csbnw, line 173:
```suggestion
    const size_t array_size = 3;
```

csbnw (commented):
Very nice addition to cudawrappers! I did a first pass over the changes and made some (minor) suggestions.

include/cudawrappers/cu.hpp
```diff
@@ -186,7 +186,7 @@ class Device : public Wrapper<CUdevice> {
   }
 
   std::string getArch() const {
-#if defined(__HIP_PLATFORM_AMD__)
+#if defined(__HIP__)
```
loostrum, line 189: Did this slip into this PR by accident? It doesn't seem related to the graph addition, and is actually wrong because `gcnArchName` is only defined on AMD GPUs, i.e. the check must be for an AMD GPU, not for the use of HIP in general.

include/cudawrappers/cu.hpp
```diff
@@ -668,6 +668,248 @@ class DeviceMemory : public Wrapper<CUdeviceptr> {
...
+    _obj.poolProps.location.id = dev.getOrdinal();
+    _obj.poolProps.location.type = CU_MEM_LOCATION_TYPE_DEVICE;
+  }
+
+  const CUdeviceptr &getDevPtr() { return _obj.dptr; }
+
+  const void *parameter() { return DeviceMemory(_obj.dptr).parameter(); }
+
+  const DeviceMemory getDeviceMemory() {
+    return DeviceMemory(_obj.dptr, _obj.bytesize);
```
wvbbreu, line 728: As an alternative approach; the `GraphDevMemAllocNodeParams` could also inherit from `cu::DeviceMemory`, especially since a graph allocation is just a different form of a device memory allocation.

include/cudawrappers/cu.hpp
```diff
@@ -668,6 +668,248 @@ class DeviceMemory : public Wrapper<CUdeviceptr> {
...
+class GraphDevMemAllocNodeParams : public Wrapper<CUDA_MEM_ALLOC_NODE_PARAMS> {
+ public:
+  GraphDevMemAllocNodeParams(const Device &dev, size_t size) {
+    _obj.bytesize = size;
+    _obj.poolProps.allocType = CU_MEM_ALLOCATION_TYPE_PINNED;
+    _obj.poolProps.location.id = dev.getOrdinal();
+    _obj.poolProps.location.type = CU_MEM_LOCATION_TYPE_DEVICE;
+  }
+
+  const CUdeviceptr &getDevPtr() { return _obj.dptr; }
```
wvbbreu, line 723: I wonder if we should hoist this method to the `cu::Device` class. In this way, you could also return a `const Device&` element. Then, other cudawrappers' graph-related functions can just accept a `cu::Device` pointer, instead of a raw `CUdeviceptr`.

include/cudawrappers/cu.hpp
```diff
@@ -668,6 +668,248 @@ class DeviceMemory : public Wrapper<CUdeviceptr> {
...
+  const void *parameter() { return DeviceMemory(_obj.dptr).parameter(); }
+
+  const DeviceMemory getDeviceMemory() {
+    return DeviceMemory(_obj.dptr, _obj.bytesize);
+  }
+};
+
+class GraphMemCopyToDeviceNodeParams : public Wrapper<CUDA_MEMCPY3D> {
+ public:
+  GraphMemCopyToDeviceNodeParams(const CUdeviceptr &dst, const void *src,
```
wvbbreu, line 734: Please refer to my earlier comment about `getDevPtr`.

include/cudawrappers/cu.hpp
```diff
@@ -886,6 +1128,14 @@ class Stream : public Wrapper<CUstream> {
   }
 #endif
 
+  void lunchGraph(CUgraphExec &graph) {
```
wvbbreu, line 1127: `lunchGraph` = `launchGraph`? ;)

matmanc: I guess I was hungry :D

matmanc: Fixed it :)

include/cudawrappers/cu.hpp
matmanc: I see what you mean but I guess you mean but maybe it is better to use the DeviceMemory one.

tests/test_graph.cpp
matmanc: Done I just set the same and cost on the top of the test

include/cudawrappers/cu.hpp
matmanc: I dont understand. Do you mean to move it outside the class?

include/cudawrappers/cu.hpp
wvbbreu: My mistake, indeed I meant `cu::DeviceMemory` as a substitution for a raw `CUdeviceptr`.

include/cudawrappers/cu.hpp
matmanc: I agree but there is the same CUdeviceptr also in the memcpyHtoD so I thought to keep it similar...

include/cudawrappers/cu.hpp
matmanc: I agree but there is the same CUdeviceptr also in the memcpyHtoD so I thought to keep it similar...

include/cudawrappers/cu.hpp
matmanc: I agree but there is the same CUdeviceptr also in the memcpyHtoD so I thought to keep it similar...

tests/test_graph.cpp
matmanc: Added :)

include/cudawrappers/cu.hpp
csbnw: You can ignore my comment, I misread the code.

include/cudawrappers/cu.hpp
matmanc: True, it can but then if you notice cu::DeviceMemory derives from Wrapper and when you construct the object so you have that both mess with _obj member and that doesnt really work well I believe

include/cudawrappers/cu.hpp
wvbbreu: We had a chat about this and decided to keep it like this.

include/cudawrappers/cu.hpp
wvbbreu: We had a chat about this and decided to keep it like this.

include/cudawrappers/cu.hpp
matmanc: So std::function is a function wrapper to get the function pointer you do `.target` but for some reason that doesnt work. Maybe I am doing it wrong....

include/cudawrappers/cu.hpp
```diff
@@ -668,6 +668,249 @@ class DeviceMemory : public Wrapper<CUdeviceptr> {
...
+
+  void addDevMemFreeNode(GraphNode &node,
+                         const std::vector<CUgraphNode> &dependencies,
+                         const CUdeviceptr &devPtr) {
+    checkCudaCall(cuGraphAddMemFreeNode(node.getNode(), _obj,
+                                        dependencies.data(),
+                                        dependencies.size(), devPtr));
+  }
+
+  void addDevMemAllocNode(GraphNode &node,
```
csbnw, line 845: Stick to the NVIDIA wording like with the other functions:
```suggestion
  void addMemAllocNode(GraphNode &node,
```

include/cudawrappers/cu.hpp
```diff
@@ -668,6 +668,249 @@ class DeviceMemory : public Wrapper<CUdeviceptr> {
...
+                          const std::vector<CUgraphNode> &dependencies,
+                          GraphDevMemAllocNodeParams &params) {
+    checkCudaCall(cuGraphAddMemAllocNode(
+        node.getNode(), _obj, dependencies.data(), dependencies.size(),
+        reinterpret_cast<CUDA_MEM_ALLOC_NODE_PARAMS *>(&params)));
+  }
+
+  void addHostToDeviceMemCopyNode(GraphNode &node,
```
csbnw, line 853: Same here:
```suggestion
  void addMemcpyNode(GraphNode &node,
```

include/cudawrappers/cu.hpp
```diff
@@ -668,6 +668,249 @@ class DeviceMemory : public Wrapper<CUdeviceptr> {
...
+    checkCudaCall(hipGraphAddMemcpyNode(
+        node.getNode(), _obj, dependencies.data(), dependencies.size(), &par_));
+
+#else
+    checkCudaCall(cuGraphAddMemcpyNode(
+        node.getNode(), _obj, dependencies.data(), dependencies.size(),
+        reinterpret_cast<CUDA_MEMCPY3D *>(&params), _context));
+#endif
+  }
+
```
csbnw, line 884: This function seems redundant, why not use the same function for both host to device and device to host copies?
```suggestion
```

include/cudawrappers/cu.hpp
```diff
@@ -668,6 +668,249 @@ class DeviceMemory : public Wrapper<CUdeviceptr> {
...
+  void exportDotFile(
+      std::string path,
+      CUgraphDebugDot_flags flags =
+          CUgraphDebugDot_flags::CU_GRAPH_DEBUG_DOT_FLAGS_VERBOSE) {
+    checkCudaCall(cuGraphDebugDotPrint(_obj, path.c_str(), flags));
+  }
+
+  CUgraphExec Instantiate(unsigned int flags = CU_GRAPH_DEFAULT) {
```
csbnw, line 892: Why the capital I? Optional: rename to `instantiateWithFlags`. (I like the shorter `instantiate`, but for consistency it may be better to stick to NVIDIA's wording.
```suggestion
  CUgraphExec instantiate(unsigned int flags = CU_GRAPH_DEFAULT) {
```

include/cudawrappers/cu.hpp
```diff
@@ -668,6 +668,249 @@ class DeviceMemory : public Wrapper<CUdeviceptr> {
...
+        node.getNode(), _obj, dependencies.data(), dependencies.size(), &par_));
+
+#else
+    checkCudaCall(cuGraphAddMemcpyNode(
+        node.getNode(), _obj, dependencies.data(), dependencies.size(),
+        reinterpret_cast<CUDA_MEMCPY3D *>(&params), _context));
+#endif
+  }
+
+  void exportDotFile(
```
csbnw, line 885:
```suggestion
  void debugDotPrint(
```

include/cudawrappers/cu.hpp
```diff
@@ -886,6 +1129,14 @@ class Stream : public Wrapper<CUstream> {
   }
 #endif
 
+  void launchGraph(CUgraphExec &graph) {
```
csbnw, line 1132:
```suggestion
  void graphLuanch(CUgraphExec &graph) {
```

tests/test_graph.cpp
```diff
@@ -0,0 +1,194 @@
...
+  nvrtc::Program program(kernel, "kernel.cu");
+  program.compile({});
+  cu::Module module(static_cast<const void*>(program.getPTX().data()));
+
+  cu::Stream stream;
+  const size_t array_size = 1000;
+  SECTION("Test cu::hostNode single value") {
+    auto fn = [](void* data) {
+      int* ptr = static_cast<int*>(data);
+      std::cout << "Host node content: " << *ptr << std::endl;
```
csbnw, line 37: Put inside `DEBUG` guards, or consider removing the print statements altogether (also the other ones below).
If you decide to do so, also remove the include of `iostream`.
```suggestion
      std::cout << "Host node content: " << *ptr << std::endl;
```

include/cudawrappers/cu.hpp
```diff
@@ -886,6 +1129,14 @@ class Stream : public Wrapper<CUstream> {
...
+    checkCudaCall(cuGraphLaunch(graph, _obj));
+  }
+
+  void launchGraph(GraphExec &graph) {
```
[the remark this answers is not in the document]
matmanc: done

tests/test_graph.cpp
```diff
@@ -0,0 +1,194 @@
+#include <array>
+#include <catch2/catch_template_test_macros.hpp>
+#include <catch2/catch_test_macros.hpp>
+#include <cstring>
+#include <iostream>
+#include <string>
+
+#include <cudawrappers/cu.hpp>
+#include <cudawrappers/nvrtc.hpp>
```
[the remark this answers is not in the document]
matmanc: Rorderdered. I think vscode was trying to be smart on me

include/cudawrappers/cu.hpp
matmanc: Changed to suggested naming

include/cudawrappers/cu.hpp
```diff
@@ -641,6 +641,249 @@ class DeviceMemory : public Wrapper<CUdeviceptr> {
...
+        node.getNode(), _obj, dependencies.data(), dependencies.size(), &par_));
+
+#else
+    checkCudaCall(cuGraphAddMemcpyNode(
+        node.getNode(), _obj, dependencies.data(), dependencies.size(),
+        reinterpret_cast<CUDA_MEMCPY3D *>(&params), _context));
+#endif
+  }
+
+  void debugDotFile(
```
csbnw, line 858:
```suggestion
  void debugDotPrint(
```

include/cudawrappers/cu.hpp
```diff
@@ -641,6 +641,249 @@ class DeviceMemory : public Wrapper<CUdeviceptr> {
...
+  }
+};
+
+class GraphMemCopyToHostNodeParams : public Wrapper<CUDA_MEMCPY3D> {
+ public:
+  GraphMemCopyToHostNodeParams(void *dst, const CUdeviceptr &src,
+                               size_t element_size, size_t size_x,
+                               size_t size_y, size_t size_z, size_t pitch = 0) {
+#if defined(__HIP__)
+    memset(&_obj, 0, sizeof(hipMemcpy3DParms));
```
csbnw, line 749: Use `std::fill` instead of `memset`?

matmanc: I am not sure this looks better
```
std::fill(&_obj, &_obj + sizeof(CUDA_MEMCPY3D), 0);
```
also is not setting an array it is just zeroing a struct.

matmanc: I found a way that looks nice and it doesnt use memset
```
_obj = {0};
```

csbnw: Very nice!

csbnw (approved):

Issue #313, opened 2025-01-29
Removal of cu::Context


wvbbreu:
In AMD HIP, the usage of GPU contexts (`hipContext`) is deprecated. Therefore, starting from cudawrappers version 0.9.0, their use is discouraged by throwing a 'deprecated' message (see issue #310). Given that all methods contained within `cu::Context` can be implemented in another scope, for instance in `cu::Device`, `cu::Context` should be removed entirely.

PR #310, opened 2025-01-20
Add deprecated message to cu::Context

wvbbreu:
Since contexts are deprecated in HIP, their use should be discouraged. This pull request introduces a deprecated message when instantiating a `cu::Context` instance.  In a future version of cudawrappers, we plan to remove `cu::Context` entirely (see issue #313).

PR #314, opened 2025-01-29
Improve HIP documentation

wvbbreu:
**Description**

This pull request improves the existing HIP-related documentation and adds a warning that the usage of `cu::Context` is discouraged.

**Related issues**:

- #310 
- #313 

- Check if the `README.md` contains the required changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ or newer|
 
 ## Usage
 
-CMake is used to build cudawrappers. To compile the library, run:
+CMake is used to build cudawrappers. To build the library, run:
 
 ```shell
 git clone https://github.com/nlesc-recruit/cudawrappers
@@ -55,7 +55,7 @@ make -C build
 This creates a `build` directory. Since cudawrappers is header-only, no library objects will be built.
 For more details on the building requirements and testing, see the [developer documentation](README.dev.md).
 
-To install cudawrappers at a specific location (e.g., ~/.local), use:
+To install cudawrappers to a specific location (e.g., ~/.local), use:
 ```shell
 cmake -DCMAKE_INSTALL_PREFIX=$HOME/.local -S . -B build
 make -C build
@@ -84,7 +84,7 @@ To enable HIP:
 
 3. Optionally, use `#ifdef (__HIP__)` directives in your source code to enable/disable certain sections for HIP.
 
-4. **Build** while ensuring that the `hipcc` compiler is selected.
+4. **Build** ensure that the `hipcc` compiler is selected.
     This can be done via the command line:
 
     ```shell
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ To enable HIP:
 
 3. Optionally, use `#ifdef (__HIP__)` directives in your source code to enable/disable certain sections for HIP.
 
-4. **Build** ensure that the `hipcc` compiler is selected.
+4. **Build**: ensure that the `hipcc` compiler is selected.
     This can be done via the command line:
 
     ```shell
--- a/README.md
+++ b/README.md
@@ -93,9 +93,8 @@ To enable HIP:
 
 **Note**: When building for both NVIDIA and AMD HIP,
 using **seperate** build folders (e.g, `build_nvidia` and `build_amd`) is encouraged.
-Additionally, please note that HIP deprecated support for so-called GPU 'contexts'.
-Therefore, the usage of cudawrappers' `cu::Context` is discouraged and marked deprecated since cudawrappers version `0.9.0`.
-
+Additionally, please note that contrary to CUDA, HIP does not support GPU 'contexts'.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -52,31 +52,46 @@ cmake -S . -B build
...
+
+3. Optionally, use `#ifdef (__HIP__)` directives in your source code to enable/disable certain sections for HIP.
+
+4. **Build** while ensuring that the `hipcc` compiler is selected. This can be done via the command line:
+
+    ```shell
+    CXX=hipcc cmake -B build
+    ```
+
+**Note**: When building for both NVIDIA and AMD HIP, using **seperate** build folders (e.g, `build_nvidia` and `build_amd`) is encouraged. Additionally, please note that HIP deprecated support for so-called GPU 'contexts'. Therefore, the usage of cudawrappers' `cu::Context` is discouraged and marked deprecated since cudawrappers version `0.9.0`.
```
matmanc, line 91: Here is says it is discouraged but context just dont work on cudawrappers when you use hip. You get a silent failure with an empty class...

matmanc:
The rests looks good to me :)

README.md
wvbbreu: Done

matmanc (approved):
Looks good to me

README.md
```diff
@@ -43,7 +43,7 @@ or newer|
 
 ## Usage
 
-We use CMake in this project, so you can clone and build this library with the following steps:
+CMake is used to build cudawrappers. To compile the library, run:
```
loostrum, line 46: As there is nothing to compile:
```suggestion
CMake is used to build cudawrappers. To build the library, run:
```

README.md
```diff
@@ -52,33 +52,53 @@ cmake -S . -B build
 make -C build
 ```
 
-This command will create a `build` directory. Cudawrappers is header only, so no library objects are being built.
-For more details on the building requirements and on testing, check the [developer documentation](README.dev.md).
+This creates a `build` directory. Since cudawrappers is header-only, no library objects will be built.
+For more details on the building requirements and testing, see the [developer documentation](README.dev.md).
 
-To install to `~/.local`, use
+To install cudawrappers at a specific location (e.g., ~/.local), use:
```
loostrum, line 58:
```suggestion
To install cudawrappers to a specific location (e.g., ~/.local), use:
```

README.md
```diff
@@ -52,33 +52,53 @@ cmake -S . -B build
...
+2. Ensure that every target your project builds is 'hipfied'.
+    This is done by marking relevant source files as 'HIP' compatible:
+
+    ```cmake
+    set_source_files_properties(source.cpp PROPERTIES LANGUAGE HIP)
+    ```
+
+3. Optionally, use `#ifdef (__HIP__)` directives in your source code to enable/disable certain sections for HIP.
+
+4. **Build** while ensuring that the `hipcc` compiler is selected.
```
loostrum, line 87:
```suggestion
4. **Build** ensure that the `hipcc` compiler is selected.
```

Note: is this still necessary at all?

README.md
```diff
@@ -52,33 +52,53 @@ cmake -S . -B build
...
-not work on non-NVIDIA GPUs, in those cases use `#ifdef(__HIP__)` guards to
-patch your code wherever this is needed.
+    ```shell
+    CXX=hipcc cmake -B build
+    ```
 
-### Usage examples
+**Note**: When building for both NVIDIA and AMD HIP,
+using **seperate** build folders (e.g, `build_nvidia` and `build_amd`) is encouraged.
+Additionally, please note that contrary to CUDA, HIP does not support GPU 'contexts'.
```
loostrum, line 96: HIP still supports it, but has it marked deprecated. Not sure it's worth changing this to reflect  that though.

loostrum (changes requested):
I happened to notice this PR, HIP always triggers my attention :) 
I've left a few minor comments.

README.md
wvbbreu: Originally, it was written in the way you are suggesting, however, Mattia pointed out that HIP basically constructs empty code blocks. Therefore, the wording was changed.

loostrum: cudawrappers does that indeed, my point was more aimed at HIP itself, which does still support contexts although they are deprecated.

loostrum: I would change "HIP" to "The HIP backend" to indicate that it relates to cudawrappers, not HIP itself. Looks good otherwise!

csbnw:
The changes look good to me, but I do have a remark.

It is possible to build with HIP without using `hipcc`. The non-HIP code can for instance be built using the system default compiler (e.g. GCC). Usually, specifying `PATH=$PATH:/opt/rocm/bin` is enough to let CMake detect the `hipconfig` binary and configure the rest. Is this a way of using HIP that we would also encourage to the users? If so, it should also be mentioned in the documentation.

Issue #306, opened 2024-11-18
Runtime Error When Mixing Runtime and Driver APIs Due to Context Management


NiclasEsser1:
**Description**

I am encountering an issue when mixing CUDA's Runtime API and Driver API within my project. While my project uses the Runtime API exclusively, I depend on an external library that uses cudawrappers and thus requires the Driver API. The issue arises when a Driver API context is created after the primary runtime context is already initialized.

When the Driver API context goes out of scope, subsequent Runtime API calls result in runtime errors. Specifically, I encounter the following error:

```
terminate called after throwing an instance of 'thrust::THRUST_200500_890_NS::system::system_error'
  what():  CUDA free failed: cudaErrorInvalidValue: invalid argument
Aborted (core dumped)
```

**Steps to Reproduce**
The following minimal reproducible example demonstrates the issue:
```
#include <cuda.h>
#include <cudawrappers/cu.hpp>
#include <thrust/device_vector.h>


int main() {
    cu::init();
    auto vec = std::make_unique<thrust::device_vector<int>>(); // First call to the runtime API -> creates primary
    auto device = std::make_unique<cu::Device>(0);
    {
        // auto context = device->primaryCtxRetain();
        cu::Context context(0, *device);
        context.setCurrent();
        vec->resize(100);
    } // context goes out of scope here
    device.reset();
    vec.reset(); // Fails with runtime error
}
``` 
**Observations**

Currently, it appears there is no way to instruct cudawrappers to utilize the primary context created by the Runtime API. However, I noticed the existence of the `cu::Device::primaryCtxRetain()` method, which seems like a potential solution. Unfortunately, it is not implemented in the current version of cudawrappers.

**Solution**
To solve the issue, I've implemented `primaryCtxRetain` as follows:
```
Context Device::primaryCtxRetain()
{
#if !defined(__HIP__)
    CUcontext primary;
    checkCudaCall(cuDevicePrimaryCtxRetain(&primary, _obj));
    return {primary, *this}; // Call to the private Context constructor -> friended to cu::Device 
#endif
}  
```
Note: The implementation must be done after the `cu::Context` class. The `primaryCtxRetain()` declaration must also be inlined in the `cu::Device' class.

With this implementation, the following adjusted example works without any errors:
```
int main() {
    cu::init();
    auto vec = std::make_unique<thrust::device_vector<int>>();
    auto device = std::make_unique<cu::Device>(0);
    {
        auto context = device->primaryCtxRetain();
[...]

PR #316, opened 2025-02-05
Remove cu::Context class

csbnw:
Remove the `cu::Context` class. The Context Management is now done in the constructor of the cu::Device class.
When a Primary Context was already active, e.g. when combining cudawrappers with the NVIDIA Runtime API, that context is retained. If not, a new Context is created.

Some tests had to be removed, while others were adapted slightly. When building in HIP mode, the Context code is disabled. The code is tested locally on NVIDIA en AMD en all pass.

**Description**

**Related issues**:

- https://github.com/nlesc-recruit/cudawrappers/issues/306
- https://github.com/nlesc-recruit/cudawrappers/issues/313

### DIFF
--- a/include/cudawrappers/cu.hpp
+++ b/include/cudawrappers/cu.hpp
@@ -53,6 +53,20 @@ inline int driverGetVersion() {
   return version;
 }
 
+size_t getFreeMemory() {
+  size_t free;
+  size_t total;
+  checkCudaCall(cuMemGetInfo(&free, &total));
+  return free;
+}
+
+size_t getTotalMemory() {
+  size_t free;
+  size_t total;
+  checkCudaCall(cuMemGetInfo(&free, &total));
+  return total;
+}
+
 inline void memcpyHtoD(CUdeviceptr dst, const void *src, size_t size) {
 #if defined(__HIP__)
   // const_cast is a temp fix for https://github.com/ROCm/ROCm/issues/2977
@@ -121,6 +135,16 @@ class Device : public Wrapper<CUdevice> {
 
   explicit Device(int ordinal) : _ordinal(ordinal) {
     checkCudaCall(cuDeviceGet(&_obj, ordinal));
+#if !defined(__HIP__)
+    unsigned int flags;
+    int active;
+    checkCudaCall(cuDevicePrimaryCtxGetState(_obj, &flags, &active));
+    if (active) {
+      checkCudaCall(cuDevicePrimaryCtxRetain(&_pctx, _obj));
+    } else {
+      checkCudaCall(cuCtxCreate(&_pctx, flags, _obj));
+    }
+#endif
   }
 
   struct CUdeviceArg {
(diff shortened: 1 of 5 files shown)


john-romein:
I do not think that I like it.  It is a major deviation from the idea to stay close to the CUDA driver API.  Also, it breaks basically all the libraries and applications that we have.

csbnw:
Going forward, I see a couple of options. In no particular order:

1. Don't merge the suggested changed
2. Keep the `cu::Context`, as is
3. Keep the `cu::Context`, but make all functions no-ops
4. Proceed with merging this code

For 1. and 2., we would also have to revert the deprecation message to improve the user experience.

I would be strongly in favour of option 3. or 4. Option 3. maintains compatibility with existing code, but could also confuse the user. People arguably shouldn't use the `main` branch and instead use any of the released versions. If they stick to an older version, API changes are not really an issue.

There is one thing that still may be need to be addressed for 3. or 4., the `cu::Context::setCurrent`. Some codes need it. In case of option 3., this won't work as there will be no `CUcontext` left. We would have to add something like `cu::Device::setCurrentContext`.

include/cudawrappers/cu.hpp
```diff
@@ -121,6 +135,16 @@ class Device : public Wrapper<CUdevice> {
...
   explicit Device(int ordinal) : _ordinal(ordinal) {
     checkCudaCall(cuDeviceGet(&_obj, ordinal));
+#if !defined(__HIP__)
+    unsigned int flags;
+    int active;
+    checkCudaCall(cuDevicePrimaryCtxGetState(_obj, &flags, &active));
+    if (active) {
+      checkCudaCall(cuDevicePrimaryCtxRetain(&_pctx, _obj));
+    } else {
+      checkCudaCall(cuCtxCreate(&_pctx, flags, _obj));
```
matmanc, line 145: If you create it here you are responsible for it so maybe you can set the the context inside a ptr and change the manager to do nothing in the first two cases

csbnw: Good point, we should indeed clean-up afterwards. I now added a `_context_manager` member that does this. It releases the context in case there was already a primary context, or it destroys it when it created one.

wvbbreu (commented):
I have a couple of minor code suggestions. I agree that the flood of deprecated messages is frustrating, so I support finding a better solution as soon as possible.

I do agree with @john-romein that these changes will have a significant impact, maybe even warranting a release bump to `1.x.x`. Based on that, I have two suggestions:

1. Delay the changes to a later version of cudawrappers and first release `0.9.0`, either with or without the deprecated messages.
2. Implement the changes now in cudawrappers version `0.9.0`, but temporarily include an empty `cu::Context` to maintain backward compatibility. This would give cudawrappers users a grace period to adapt.

include/cudawrappers/cu.hpp
```diff
@@ -208,134 +230,10 @@ class Device : public Wrapper<CUdevice> {
   int getOrdinal() const { return _ordinal; }
 
  private:
+  std::shared_ptr<CUcontext> _context_manager;
   int _ordinal;
 };
 
-class Context : public Wrapper<CUcontext> {
```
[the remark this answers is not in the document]
csbnw: I don't know if anyone is really using these functions. They are not available with HIP, are they? I think it's sufficient to inform the user about `cu::Context` being removed in the changelog.

Issue #327, opened 2025-03-25
Graph should not use hipContext


loostrum:
When compiling in HIP mode, deprecation warnings related to `hipContext` show up even though `cu::Context` was made such that it is effectively no-op for HIP. I found that the reason is that `cu::Graph` uses `cuContext` / `hipContext` directly, instead of the wrapped cudawrappers version. This makes the `hipContext` deprecation warnings show always when including `cu.hpp`. I'm not 100% sure this classifies as a bug as there is no functional problem, but we did want to avoid the big wall of deprecation warnings with HIP which is now back.

A solution may be to use a `cu::Context` object inside `cu::Graph`, instead of the direct CUDA/HIP objects.

PR #328, opened 2025-03-25
Use cu::Context in cu::Graph

loostrum:
Use `cu::Context` in  `cu::Graph` instead of directly accessing `CUDA` / `HIP` context functions. This avoids using HIP contexts while these are deprecated, see issue #327. 

This is an API-breaking change: a `cu::Context` object now needs to be passed to the `cu::Graph` constructor.

I ran the tests on both an NVIDIA and AMD GPU and all pass.

Note: `cu::Graph` is not mentioned in the changelog at all, so I thought it would be strange to mention changes to it in this PR. Perhaps a note of the addition of graph support should be added in a separate PR?

### DIFF
--- a/include/cudawrappers/cu.hpp
+++ b/include/cudawrappers/cu.hpp
@@ -790,17 +790,16 @@ class GraphMemCopyToHostNodeParams : public Wrapper<CUDA_MEMCPY3D> {
 
 class Graph : public Wrapper<CUgraph> {
  public:
-  explicit Graph(CUgraph &graph) : Wrapper(graph) {
-    checkCudaCall(cuCtxGetCurrent(&_context));
-  };
+  explicit Graph(cu::Context &context, CUgraph &graph)
+      : Wrapper(graph), _context(context) {};
 
-  explicit Graph(unsigned int flags = CU_GRAPH_DEFAULT) {
+  explicit Graph(cu::Context &context, unsigned int flags = CU_GRAPH_DEFAULT)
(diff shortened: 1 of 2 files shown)


include/cudawrappers/cu.hpp
```diff
@@ -790,17 +790,16 @@ class GraphMemCopyToHostNodeParams : public Wrapper<CUDA_MEMCPY3D> {
...
  public:
-  explicit Graph(CUgraph &graph) : Wrapper(graph) {
-    checkCudaCall(cuCtxGetCurrent(&_context));
-  };
+  explicit Graph(cu::Context &context, CUgraph &graph)
+      : Wrapper(graph), _context(context) {};
 
-  explicit Graph(unsigned int flags = CU_GRAPH_DEFAULT) {
+  explicit Graph(cu::Context &context, unsigned int flags = CU_GRAPH_DEFAULT)
+      : _context(context) {
```
loostrum, line 797: Note sure if this should be `_context(context.getCurrent())` instead.

csbnw:
With [these](https://gitlab.com/astron-misc/cudapeak/-/commit/c0cfcaaccaa7af5bb7fb6e660ffa1fadf8218cd9) changes, @matmanc's CUDA graph code works.

include/cudawrappers/cu.hpp
csbnw: `_context(context)` looks correct to me: just pass the context along.

PR #329, opened 2025-03-26
Add mangled name support

### LINKED ISSUE #281: Mangled name support
**Is your feature request related to a problem? Please describe.**
Currently kernel names are passed directly to NVRTC. This means that C++ mangled names are not supported and kernels must have `extern "C"` linkage. However, this means templated kernels are not supported.

**Describe the solution you'd like**
Support mangled names in the cudawrappers NVRTC module.

**Describe alternatives you've considered**

**Additional context**
We can implement this with a combination of [nvrtcAddNameExpression](https://docs.nvidia.com/cuda/nvrtc/index.html#group__compilation_1ga0b07e4173b28a10682f21edc7140844e) and [nvrtcGetLoweredName](https://docs.nvidia.com/cuda/nvrtc/index.html#group__compilation_1ga2543158bd0e25649254a9d86a81e4f50)

loostrum:
**Description**

Add support for `nvrtcAddNameExpression` and `nvrtcGetLoweredName`, and their hiprtc equivalents. I think it might be nice to hide this complexity from the user entirely, but that would be quite a bit more work as the user typically gives the function name only to `cu::Function`, which knows nothing about `nvrtc`.

I didn't add an explicit test for the new functionality, but instead removed `extern "C"` from all kernels in the tests. This causes the compiler to mangle the names which is then handled by the new functionality.

**Related issues**:
Closes #281 

- Run tests on with both CUDA and HIP

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 All notable changes to this project will be documented in this file. This
 project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Added
+
+- Added `nvrtc::addNameExpression` and `nvrtc::getLoweredName`
+
 ## [0.9.0] - 2025-03-18
 
 ### Added
--- a/include/cudawrappers/nvrtc.hpp
+++ b/include/cudawrappers/nvrtc.hpp
@@ -152,6 +152,16 @@ class Program {
     return log;
   }
 
+  void addNameExpression(const std::string &name) {
+    checkNvrtcCall(nvrtcAddNameExpression(program, name.c_str()));
+  }
+
+  const char *getLoweredName(const std::string &name) {
+    const char *lowered_name;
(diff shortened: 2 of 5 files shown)


csbnw:
If I understand correctly, you can now either:

1. Use `extern "C"` as before
2. Use `nvrtcAddNameExpression` and `nvrtcGetLoweredName`

The second option doesn't seem very user-friendly, but hiding this functionality internally goes against the principle of 'just' providing the CUDA API to the user. Still, I think that handling it internally is the cleanest solution.

With two different ways to do runtime compilation, I think it is valuable to have test cases for both and would encourage you to add a test.

loostrum:
Indeed both ways now work. I'll add a test to ensure we test both ways. 
Let's discuss further how we could handle the lowered names internally.

csbnw (approved):

PR #331, opened 2025-04-25
include `cudawrappers-inline-local-includes.cmake` in the installation

infinitron:
**Description**

The current build does not install the cmake file, `cmake/cudawrappers-inline-local-includes.cmake`. This leads to build errors when `target_embed_source` function is used, which adds a dependency to this file. 
https://github.com/nlesc-recruit/cudawrappers/blob/adecd500fb803021aa11ef0b200ad1ebbdc47591/cmake/cudawrappers-helper.cmake#L42-L48
This PR fixes the `CMakeLists.txt` to include this cmake file in the install directory.

**Related issues**:

- None

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -113,10 +113,12 @@ if(CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
   )
 
   install(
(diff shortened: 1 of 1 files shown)


csbnw (approved):
We typically use `FetchContent` to include cudawrappers in other projects, the `make install` route is not well tested. Great that you caught this omission!

PR #332, opened 2025-04-29
Fix cu::DeviceMemory constructors with CUdeviceptr

csbnw:
**Description**

As identified [here](https://git.astron.nl/RD/recruit/ccglib/-/merge_requests/74#note_116140), the `cu::DeviceMemory` constructor that accept a `CUdeviceptr` cause issues. There are two problems:
- The `_obj` member is not initialized to `ptr` as it should.
- The `manager` is initialized to `cu::Wrapper`, while the `ptr` should not be managed at all.

The problem is solved by explicitly initializing `_obj` and by initializing `manager`  with an empty deleter.

**Related issues**:

- N.A.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 All notable changes to this project will be documented in this file. This
 project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Changed
+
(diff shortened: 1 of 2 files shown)


loostrum (approved):
Thanks for fixing this! I verified that this does work as expected in ccglib.

PR #337, opened 2025-05-21
Make inlining local includes recursive

### LINKED ISSUE #336: Make `inline_local_includes` recursive
**Is your feature request related to a problem? Please describe.**
If a local header includes another local header, the nested one doesn't get inlined and could let to HIPRTC/NVRTC errors.

**Describe the solution you'd like**
Make `inline_local_inludes` recursive.

**Describe alternatives you've considered**
Setting the include path of the runtime compiler fixes this too, but the whole point of inlining the headers and embedding the kernels source is to avoid having to do that.

**Additional context**
I ran into this in ccglib, which uses a few small headers to set types etc. Using these as part of larger headers is currently not possible if those are included in kernels as they can't be found at runtime.

loostrum:
**Description**

Both `get_local_includes` and `inline_local_includes` were adapted to call themselves for each local header found, hence enabling recursive local includes to be fully inlined. I've added tests for both a single and double include.

Note: It will fail when header A includes header B and header B includes header A. I tried a few things to avoid this, but haven't found a workaround yet. 

**Related issues**:
Closes #336

### DIFF
--- a/cmake/cudawrappers-helper.cmake
+++ b/cmake/cudawrappers-helper.cmake
@@ -1,6 +1,6 @@
 # Return a list of absolute file names for all the local includes of the
 # input_file.  Only files in the root directory will be considered.
-function(get_local_includes input_file root_dir parent_include_files)
+function(get_local_includes input_file root_dir)
   file(READ ${input_file} input_file_contents)
   set(include_regex "(^|\r?\n)(#include[ \t]*\"([^\"]+)\")")
   string(REGEX MATCHALL ${include_regex} includes ${input_file_contents})
@@ -17,10 +17,6 @@ function(get_local_includes input_file root_dir parent_include_files)
       list(APPEND include_files ${include_PATH})
     endif()
   endforeach()
-  set(${parent_include_files}
-      ${include_files}
-      PARENT_SCOPE
-  )
 endfunction()
 
 # Make it possible to embed a source file in a library, and link it to a target.


loostrum:
> What about the case when A includes B and C and B includes C. Will C be added twice? How about checking whether `${include}` is present in `${parent_include_files}`? I think this might solve both problems at once.

I played with the `IN_LIST` check but couldn't get that to work yet. I'll have a look again, indeed I think that C would be added twice in the case you describe.

tests/test_nvrtc.cpp
```diff
@@ -54,7 +54,35 @@ TEST_CASE("Test nvrtc::Program embedded source", "[program]") {
...
+}
+
+extern const char _binary_tests_kernels_recursive_include_kernel_cu_start,
+    _binary_tests_kernels_recursive_include_kernel_cu_end;
+
+TEST_CASE("Test nvrtc::Program recursively inlined header", "[program]") {
+  const std::string kernel(
+      &_binary_tests_kernels_recursive_include_kernel_cu_start,
+      &_binary_tests_kernels_recursive_include_kernel_cu_end);
+  nvrtc::Program program(kernel, "ector_add_recursive_include_kernel.cu");
```
csbnw, line 78:
```suggestion
  nvrtc::Program program(kernel, "recursive_include_kernel.cu");
```

csbnw:
After some testing, I can confirm that this setup won't prevent headers from being included twice. With just `#define`'s, this is not problematic, but it is when for instance certain helper functions are included more than once. This will lead to a `device kernel image is invalid error`.

Moreover, I noticed that a change in any of the recursively included header files won't trigger inlining+embedding.

loostrum:
@csbnw Can you share an example where the rebuild on change doesn't work? It does seem to work for me:
```
leon@zeus:~/src/cudawrappers/build$ touch ../tests/kernels/secondary.h 
leon@zeus:~/src/cudawrappers/build$ make
[ 82%] Built target Catch2
[ 83%] Built target Catch2WithMain
[ 85%] Built target test_cu
[ 88%] Built target vector_add_kernel
[ 89%] Inlining all includes of kernels/single_include_kernel.cu
[ 89%] Creating object file for kernels/single_include_kernel.cu
[ 90%] Linking CXX static library libsingle_include_kernel.a
[ 90%] Built target single_include_kernel
[ 91%] Inlining all includes of kernels/recursive_include_kernel.cu
[ 92%] Creating object file for kernels/recursive_include_kernel.cu
[ 93%] Linking CXX static library librecursive_include_kernel.a
[ 93%] Built target recursive_include_kernel
[ 94%] Linking HIP executable test_nvrtc
[ 94%] Built target test_nvrtc
[ 95%] Built target test_cufft
[ 96%] Built target test_vector_add
[ 98%] Built target test_graph
[100%] Built target test_nvtx
leon@zeus:~/src/cudawrappers/build$ touch ../tests/kernels/primary.h 
leon@zeus:~/src/cudawrappers/build$ make
[ 82%] Built target Catch2
[ 83%] Built target Catch2WithMain
[ 85%] Built target test_cu
[ 88%] Built target vector_add_kernel
[ 90%] Built target single_include_kernel
[ 91%] Inlining all includes of kernels/recursive_include_kernel.cu
[ 92%] Creating object file for kernels/recursive_include_kernel.cu
[ 93%] Linking CXX static library librecursive_include_kernel.a
[ 93%] Built target recursive_include_kernel
[ 94%] Linking HIP executable test_nvrtc
[ 94%] Built target test_nvrtc
[ 95%] Built target test_cufft
[ 96%] Built target test_vector_add
[ 98%] Built target test_graph
[100%] Built target test_nvtx
```

The `single_include` file only includes `secondary.h`, while `recursive_include` includes `primary.h` which includes `secondary.h`, so I think that above is the correct behaviour.

csbnw:
I just pushed a commit that fixes the duplicated includes. The caveat that you may also have encountered is that appending to a list does not update the parent scope, see also [here](https://cmake.org/cmake/help/latest/command/list.html#introduction).

csbnw:
> @csbnw Can you share an example where the rebuild on change doesn't work? It does seem to work for me:
You are right, it now also works for me. I must have mixed a few things up during testing.
Looking at the `include_files` variable in `cudawrappers-helper.cmake`, it indeed contains all the correct header files. Since this variable is passed to `DEPENDS`, CMake will automatically trigger the custom command when any of them change.

cmake/cudawrappers-inline-local-includes.cmake
```diff
@@ -16,14 +24,35 @@ function(inline_local_includes input_file output_file root_dir)
...
+        # Replace the include line with the include file contents
+        string(REPLACE "${include_line}" "${include_contents}"
+                       input_file_contents "${input_file_contents}"
+        )
+      else()
+        # Remove the include line
+        string(REPLACE "${include_line}" "" input_file_contents
+                       "${input_file_contents}"
+        )
+      endif()
```
loostrum, line 41: Not sure if it's good for readability, but this could be shortened a bit by setting `include_contents` to an empty string before the if statement:

```suggestion
      set(include_contents "")
      if(found_index EQUAL -1)
        list(APPEND ALREADY_INCLUDED ${include_PATH})
        inline_local_includes(${include_PATH} include_contents ${root_dir})
      endif()
      # Replace the include line with the include file contents or
      # remove it if the file was already included
      string(REPLACE "${include_line}" "${include_contents}"
                     input_file_contents "${input_file_contents}"
      )
```

csbnw: I like it! :+1:

csbnw (approved):

PR #354, opened 2025-11-04
Workaround libnvrtc-builtins.so not found at runtime

csbnw:
**Description**

Preload `libnvrtc-builtins.so` in `nvrtc::Program` constructor:
- Guarantees the built-ins library is loaded before any NVRTC API calls
- Prevents `NVRTC_ERROR_BUILTIN_OPERATION_FAILURE` on systems where `LD_LIBRARY_PATH` does not include the CUDA library directory
- Users can now use runtime compilation without manually setting LD_LIBRARY_PATH

The issue arises because NVRTC uses a separate `libnvrtc‑builtins.so` which it loads at runtime via `dlopen` by `libnvrtc.so`. If that library isn’t found in a loader‐searchable path, NVRTC fails with `NVRTC_ERROR_BUILTIN_OPERATION_FAILURE`. This has been known to surface in CUDA 11.x and later. By explicitly preloading the unversioned `libnvrtc‑builtins.so`, we mitigate the dependency on external loader paths. This cannot be fully fixed at build time via CMake (e.g., linking to `CUDA::nvrtc-builtins` added in CMake 3.26) because NVRTC internally calls `dlopen` at runtime, which ignores the executable’s `RPATH`. The library must be loaded before any NVRTC API calls to guarantee runtime compilation succeeds.

**Related issues**:

- N.A.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,60 +5,74 @@ on:
   workflow_dispatch:
 
 jobs:
-  linting:
-    name: run
-    runs-on: ubuntu-22.04
-    env:
-        cuda-version: "12.6.1"
-        hip-version: "6.1.3"
-        amdgpu-version: "6.1.60103-1_all"
+  build-cuda:
+    name: build (CUDA ${{ matrix.cuda-version }})
+    runs-on: ubuntu-24.04
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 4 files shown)


wvbbreu (changes requested):
Very nice diagnosis! I have one small remark.

wvbbreu (approved):

PR #361, opened 2025-12-05
Fix nvrtc include path for CUDA 13

wvbbreu:
**Description**

From CUDA version 13, some folders inside `/include/cuda` have been relocated to `/include/cuda/cccl`. To ensure that packages depending on this library do not break (for instance `<cuda/pipeline>`), the CCCL path will also be appended.
In the future, the `nvrtc::findIncludePath` should be removed in favor of `nvrtc::findIncludePaths`.

**Related issues**:

- #362

### DIFF
--- a/include/cudawrappers/nvrtc.hpp
+++ b/include/cudawrappers/nvrtc.hpp
@@ -40,54 +40,6 @@ inline void checkNvrtcCall(nvrtcResult result) {
   if (result != NVRTC_SUCCESS) throw Error(result);
 }
 
-inline std::string findIncludePath() {
-  std::string path;
-
-  if (dl_iterate_phdr(
-          [](struct dl_phdr_info *info, size_t, void *arg) -> int {
-            std::string &path = *static_cast<std::string *>(arg);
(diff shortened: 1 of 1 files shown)


include/cudawrappers/nvrtc.hpp
csbnw: There is now a lot of duplicated code in `findIncludePath` and `findIncludePaths`. Can't we just return the first element from whatever `findIncludePath` returns here?

wvbbreu: Fair suggestion. I thought about this already when implementing the code. However, I do think this has two complications:
1. `findIncludePath` already appends the `cccl` path, therefore should be popped again afterward. Alternatively, we would have to expose some sort of variable to control this (instead of using the preprocessor directives). This is of course possible, however, requires splitting the implementation to a private helper function.
2. Having to do some path scanning twice, thus redoing some of the work.

What do you think? I think we should not overcomplicate things here as `findIncludePath` will eventually be removed.

csbnw (approved):

